PR #3, opened 2019-04-14
Execute OpenHospital GUI with Docker compose

pviotti:
This PR adds the OpenHospital Swing-based GUI to the Docker Compose file, so that one can run both database and GUI with a simple `docker-compose up`.  

Notes:
 - I had to modify `startup.sh` because it wasn't working - several folders of those included in the Java classpath do not exist, and the one created by the Maven build (`/target/*`) are not present
 - I only tested this on Linux (and I suspect there will be issues at using this on Windows since it uses X11 for graphics..)
 - the Docker file for the GUI downloads and builds both core and gui with Maven. I guess it would be quicker to build if we published the Maven artifacts somewhere.
 - running GUI applications in Docker is less than trivial, as you have to use a non-root user, and somehow map the X11 server configuration in the container to a local one on the host machine (see for instance [this](https://medium.com/@learnwell/how-to-dockerize-a-java-gui-application-bce560abf62a))
 - the Docker image for the GUI is published [here](https://cloud.docker.com/repository/docker/pviotti/openhospital) as pviotti/openhospital, and can be run separately (tested only on Linux) with this command: `docker run -ti --rm -v /tmp/.X11-unix:/tmp/.X11-unix -v /home/$(whoami)/.Xauthority:/home/ohuser/.Xauthority:ro -e DISPLAY --network=host pviotti/openhospital:latest`
 - in the Docker file I used my fork of this repo since the `startup.sh` file does not work on the main repo

Related issues: [OP-68](https://openhospital.atlassian.net/projects/OP/issues/OP-68?filter=allopenissues) (and child taks)

### DIFF
--- a/Dockerfile-oh
+++ b/Dockerfile-oh
@@ -1,0 +1,35 @@
+FROM maven:3.6.0-jdk-11
+
+ENV OH_CORE_REPO https://github.com/informatici/openhospital-core.git
+ENV OH_CORE_BRANCH master
+
+ENV OH_GUI_REPO https://github.com/pviotti/openhospital-gui.git
+ENV OH_GUI_BRANCH startup-script
+
+# Running GUI application as root (Docker's default) is problematic
+# see https://unix.stackexchange.com/questions/118811/why-cant-i-run-gui-apps-from-root-no-protocol-specified#answer-118826
+RUN /usr/sbin/useradd --comment ohuser --home-dir /home/ohuser \
+    --non-unique --uid 1000 --user-group --system \
+    --shell /bin/bash ohuser \
+    && mkdir -p /home/ohuser \
+    && chown -R ohuser:ohuser /home/ohuser \
+    && chown -R ohuser:ohuser /opt
+USER ohuser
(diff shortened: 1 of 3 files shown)


mwithi:
Hi Paolo, I have used Xming as X server for Windows and introduced the ENV varaible DISPLAY=<my_ip>:0. Nevertheless it shows me a problem of DB connection:
http://www.pastebin.it/5781/

DB connection failed

pviotti:
It looks like the graphic server (Xming) is fine, as it manages to show the Swing dialog box. The error is about the connection with the DB, which should be listening on port 3306. So if you just started the GUI container then the error is normal - otherwise, if you started also the db container (for instance with `docker-compose up`) there's some networking issue.

mwithi:
Ok, I thought your container on dockerhub was all-in-one. Now I've started the DB with `docker-compose up` and changed docker-compose.yml with:
```
environment:
     - DISPLAY=192.168.1.8:0
```
done

It seems slightly slow but definitely we can say that at least it works!

PR #38, opened 2019-11-08
OP-126 list box

aalsabag:
This is to fix bug OP-126.
The issue only happens in java 1.6. In 1.8 it works without intervention.
I have now fixed the issue for 1.6 as per the suggestion here:
https://stackoverflow.com/questions/455753/jtextarea-very-small-size-with-long-text/455777
image

I have also confirmed that the newly added code still works in java 1.8.

Please


aalsabag:
> is that really needed to repeat pack() method?

No. I just tested this without both pack() method calls and it works in both Java 6 and 8. PR updated. Thank you for this catch.

In case you are curios. The reason I originally had this was because of what i found in this post. There is a bug in Java returning improper preferred size and this was the recommended solution:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4924163

This is the stack overflow page where I found people with the same issue
https://stackoverflow.com/questions/455753/jtextarea-very-small-size-with-long-text/455777
>  pack();
>  _definition.setSize(_definition.getPreferredSize());
>  pack();

mwithi:
Thanks very @aalsabag ! Very interesting indeed!

PR #45, opened 2020-01-04
OP-143: "patient profession" and "marital status" fields

aalsabag:
These are the necessary UI changes for OP-143 . There is a PR for openhospital-core as well that corresponds with this. Please let me know if there are any necessary changes.
Addition of profession and marital status fields.
Both are now drop downs.
Added the arabic/english translations.
Current list of professions to be expanded.
Updated the summary page to show these new fields
Included DB scripts (even though they are included in openhospital-core)

image
image
image


aalsabag:
> Hi @aalsabag, well done!
> 
> For this and future enhancements of this kind, where we have `dictionary fields` that need to be translated, we should put in the DB only a coded key and to translate it only when it is showed to the screen or in the reports (throught a similar bundle.properties). In other words in the DB cannot go a translated string but only code-awared codified values.
> 
> Is it possible for you to develop this behavior like has been done in these other PR for lab materials and examination parameters?
> [informatici/openhospital-core#42](https://github.com/informatici/openhospital-core/pull/42)
> [informatici/openhospital-core#43](https://github.com/informatici/openhospital-core/pull/43)
> 
> **PS**: this should be applied also for old features that don't follow this pattern

Thanks for the feedback! I believe I undestand. Just to clarify, you mean that the DB should not store the word "Widowed", for example, but rather "angal.patient.maritalstatuswidowed". Then I should just translate upon displaying. Do I have the correct understanding here?

mwithi:
Hi @aalsabag, rather than bundle key I'm suggesting to use self-explaining keys, like "widowed" (lowecase) and map it programmatically in the "gui" with "widowed -> angal.patient.maritalstatuswidowed". 

In this way, looking at the tables, we have something human-readable, and translations is shifted to the "gui" and eventually to the Jasper report with a <report_file_name>_<code>.properties file as already done for some reports.

What do you think?

aalsabag:
Yup this makes sense. This is what I did for the marital status [link](https://github.com/informatici/openhospital-gui/pull/45/files#diff-56fac2d45ee6208740bbacb3e952f089R1116-R1133). I build a hashmap that contains a key/value pair. The key is what is stored in the DB. Thanks for the clarification. My only question though. You mentioned to do the mapping in the gui, but don't you think the logic is better suited for openhospital-core instead of openhospital-gui? That way I can reference the key/value getters in multiple places. If you don't think that is best I can just leave it here and add the commit for the profession field, otherwise I will move this to the Patient class in openhospital-core. Your call :)

mwithi:
Yes, sorry for the mispelling, it has to be done at the manager level, that is actually in the core, like the two pull-requests I've initially linked as an example.

Cheers! 🙌

aalsabag:
> Yes, sorry for the mispelling, it has to be done at the manager level, that is actually in the core, like the two pull-requests I've initially linked as an example.
> 
> Cheers! 🙌

All done here. Let me know if anything should be changed/added. I have a few days off now :)

PR #70, opened 2020-03-11
Create language_zh_CN.properties

feilong1976:


mwithi:
Starting from now, the same translation could be found and operated on Transifex ([link](https://www.transifex.com/informatici-senza-frontiere-onlus/openhospital/language/zh_CN/)) along side with other project resources.

PR #73, opened 2020-03-13
OP-105-string_to_date_in_jasper_reports

estromsnes:
I have modified the reports to support date parameters as java.util.Date objects (instead of String).

The parameter data types have been changed, the sql statements have been modified to support pure dates and some formatting has been added to the date fields to ensure correct formatting of the dates.

This pull request is dependent on an equal pull-request in the openhospital-core project where I have modified the data type of date parameters that is passed to the Jasper engine when creating reports.


mwithi:
Hi @estromsnes, since we have new a new layout in the code and also a new default branch, is it possibile to rebase the work onto `develop`?

PR #76, opened 2020-03-22
OP-158: Prevent GUI freeze when opening some window due to deadlock

alexwibowo:
* Thread 1 (looks like it is the EDT): CursorManager was locking the WaitCursorEventQueue, and then attempt to hold lock on underlying EventQueue
* Thread 2: has lock on the underlying EventQueue tried to hold lock on WaitCursorEventQueue
* Tried to fix this by making sure Thread 1 acquire the underlying EventQueue first, before holding lock on WaitCursorEventQueue (in the same order as Thread 2)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -285,6 +285,11 @@
 	  		<artifactId>spring-context</artifactId>
 	  		<version>${spring.framework.version}</version>
 	  	</dependency>
+		<dependency>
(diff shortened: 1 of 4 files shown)


mwithi:
Superseded by https://github.com/informatici/openhospital-gui/pull/77

PR #98, opened 2020-05-03
OP-73 using new patient profile table

alexwibowo:


mwithi:
Hi @alexwibowo, since we have new a new layout in the code and also a new default branch, is it possibile to rebase the work onto `develop`?

PR #100, opened 2020-05-10
OP-99 #comment Print payments before closing bill's window #time 2h #…

fjevictoriano:
With these changes the system will check if there are new payments added and it will ask to print the payments after saving them and before closing the bill's window.

### DIFF
--- a/src/org/isf/accounting/gui/PatientBillEdit.java
+++ b/src/org/isf/accounting/gui/PatientBillEdit.java
@@ -1144,6 +1144,11 @@ public class PatientBillEdit extends JDialog implements SelectionListener {
 						fireBillInserted(updateBill);
 						
 					}
+					if (hasNewPayments()) {
+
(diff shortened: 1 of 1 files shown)


src/org/isf/accounting/gui/PatientBillEdit.java
```diff
@@ -1156,7 +1161,11 @@ public void actionPerformed(ActionEvent e) {
 		}
 		return jButtonSave;
 	}
-	
+
+	private boolean hasNewPayments() {
+		return insert && payItems.size() > 0 || (payItems.size() - payItemsSaved) > 0;
```
tomek39856, line 1166: (insert && payItems.size() > 0) for better readability?

fjevictoriano: Sure! thanks

PR #156, opened 2020-05-31
OP-181 Standard Maven Layout

mwithi:
Moved files to comply Standard Maven Layout

**To Check**:
- [x] removed mysql/ folder (only needed for core) but it should included in distribution
- [x] should rsc/ folder remain for user configurables settings? 
- [x] included OP-216
- [ ] ~~should include https://github.com/informatici/openhospital/issues/14 ?~~


pom.xml
```diff
@@ -53,6 +44,7 @@
 							<outputDirectory>${basedir}/target/OpenHospital20/rsc</outputDirectory>
 							<resources>
 								<resource>
+									<directory>src/main/resources</directory>
```
pviotti, line 47: I think this is the [standard location](http://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) and therefore can be omitted (but I might be wrong, haven't tested it).

mwithi:
> it is indeed removed - and it doesn't show up in `target/OpenHospital20` (is this OK? aren't we adding it only in the openhospital root project before packaging it?)

Yes, it is removed and we should not need it in the target folder (since we have the docker-composer)
Yes, we need to add in openhospital root project before packaging it, but we were taking from "gui" if I'm not wrong, so we should modify the build_poh.sh script when we are ager for the next release.

> Don't know what's the goal of those files and whether any user every actually touched them.
In doubt, for precaution, I think you can leave it there.

Yes, they are needed because those are files that can be changed by the user as described in the AdminManual#5-configuration-files)

> not sure. We can do this in another PR. It's already a big change to just move the code to the standard folders.

Yes, I agree...

pom.xml
mwithi: this is in the `<goal>copy-resources</goal>` so it is needed to build the target I think and not for Maven itself

src/main/java/org/isf/medicalstock/gui/MovStockMultipleDischarging.java
```diff
@@ -94,7 +94,7 @@
 	 */
 	public static void main(String[] args) {
 		try {
-			PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath()); //$NON-NLS-1$
+			PropertyConfigurator.configure(new File("./src/main/resources/log4j.properties").getAbsolutePath()); //$NON-NLS-1$
```
tomek39856, line 97: you can try to read it from classpath instead of being dependent on project structure, sth like this: https://howtodoinjava.com/java/io/read-file-from-resources-folder/

src/main/java/org/isf/sms/gui/SmsBrowser.java
```diff
@@ -405,7 +405,7 @@ public String formatTodayDateTime(Date smsDateSent) {
 	public static void main(String[] args) {
 		GeneralData.getGeneralData();
 		MessageBundle.initialize();
-		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath()); //$NON-NLS-1$
+		PropertyConfigurator.configure(new File("./src/main/resources/log4j.properties").getAbsolutePath()); //$NON-NLS-1$
```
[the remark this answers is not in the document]
mwithi: This is a main method for testing indeed, so no need to focus to much on it, but in general we have the same issue in other parts of the code, we should think on it.

src/main/java/org/isf/medicalstock/gui/MovStockMultipleDischarging.java
mwithi: This is a main method for testing purposes, so no need to focus to much on it, but in general we have the same issue in other parts of the code, we should create an issue to fix them.

PR #183, opened 2020-07-09
[OP-226] Java 8

pviotti:
About the issue with the `dpiaware` flag mentioned in [OP-226](https://openhospital.atlassian.net/browse/OP-226), I can't reproduce it since I don't own a high-dpi screen. And I'd suggest to survey whether this is a big issue with OH users.  
Anyway, there are some (awkward) [solutions out there](https://stackoverflow.com/a/39372897/153761), and apparently [the issue is solved in Java 9](https://stackoverflow.com/a/45426996/153761).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,8 +26,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.6</source>
-					<target>1.6</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>


mwithi:
Personally I've tried all the proposed solutions but no one worked (some old-fashioned Java programs seems work but the final effect is blurry and not very nice), also it would be difficult to document for the end-users. I think high-dpi monitors will be available in the future but we cannot be blocked with this issue, so I agree the priority is low. Bumping to Java9 seems too big step at this time.

Nevertheless I don't think that changing the compiler to Java8 would be called 'porting', but as you said [here](https://github.com/informatici/openhospital-core/pull/46#pullrequestreview-446029788) we can do it gradually.

Nice todo list [here](https://corejava25hours.com/2018/03/23/java-7-vs-java-8-vs-java-9-vs-java-10/)

mwithi:
Tried again [this solution](https://stackoverflow.com/a/39372897/153761) and worked.

I think we can merge and leave next changes to future PRs

PR #195, opened 2020-08-18
Update imports

dbmalkovsky:
Per the comment in this [PR](https://github.com/informatici/openhospital-core/pull/149#issuecomment-675482816) imports in this project are updated to:

- remove unused imports
- order the imports in a common order across repositories
- replace package imports with individual import statements


src/main/java/org/isf/sms/gui/SmsBrowser.java
```diff
@@ -38,12 +38,11 @@
...
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.gui.OHServiceExceptionUtil;
+import org.isf.utils.jobjects.CustomJDateChooser;
 import org.isf.utils.jobjects.ModalJFrame;
 import org.joda.time.DateMidnight;
 import org.joda.time.DateTime;
 
-import org.isf.utils.jobjects.CustomJDateChooser;
-
 /**
```
pviotti, line 46: not related to this PR, but I'd be in favor of getting rid of these custom @author headers, and maybe replace them with some predefined license/copyright ones (there's even a [maven plugin](https://www.mojohaus.org/license-maven-plugin/) to do that automatically)

pviotti (approved):
looks good, thanks. Next step would be adding the ide-settings folder and documenting a bit the process of loading those settings, as in the core project.

src/main/java/org/isf/sms/gui/SmsBrowser.java
dbmalkovsky: See this [PR](https://github.com/informatici/openhospital-core/pull/165) which has an example of using the maven plugin.

PR #198, opened 2020-08-20
Replace concatenation with parameterized log message

dbmalkovsky:
Non-constant concatenations in logging messages are evaluated at runtime even when the logging message is not logged; this can negatively impact performance. Thus these changes use a parameterized log message instead which will not be evaluated when logging is disabled.

Because there were only 9 files, reformatting to match the standards was done.

If you want to hide the white space changes:
HideWhiteSpaceChanges


mwithi (approved):
LGTM, anyway does the formatter we are adopting have a limit on 80char?

dbmalkovsky:
@mwithi No, the formatting setting has a hard line limit at 160 characters.

PR #202, opened 2020-08-30
[OP-229] disabled auto creating profile photo relation, fixed patient search

tomek39856:


mwithi (approved):
It works fine together with https://github.com/informatici/openhospital-core/pull/188.

PR #203, opened 2020-08-31
Transifex badges in README.md (WIP)

mwithi:
~Added transifex badges using https://transifex-open-api.herokuapp.com/~

**Update**: the above app is no longer available. To evaluate using a GitHub action to query Transifex API and collect the needed information and calculate the translation percentages (not provided directly)

e.g. https://transifex.github.io/openapi/index.html#tag/Statistics/paths/~1resource_language_stats/get


mwithi:
The final effect it is not the best one... I think it should be possible to filter the translation progresses by resources categories (application or reports, at the moment) and translation treshold (e.g. above 80%) that is also what triggers the PR from Transifex

PR #207, opened 2020-09-04
OP-104 upgrade jasper report library

mwithi:
This PR is taking care of #205

The same as https://github.com/informatici/openhospital-core/pull/198 is doing for https://github.com/informatici/openhospital-core/pull/123

New Jasper version is 6.14.0, all reports have been updated.

*DONE*: To update openhospital-doc/ any reference to Jasper
https://github.com/informatici/openhospital-doc/pull/26


pviotti:
1. any hint on how to test these changes and the one in https://github.com/informatici/openhospital-core/pull/198 ? Should we just try to produce some reports?
2. why are .bak and .old files checked in?
3. should both jrxml and jasper files be checked in? IIRC, one is the compiled version of the other, right..?

mwithi:
> 1. any hint on how to test these changes and the one in [informatici/openhospital-core#198](https://github.com/informatici/openhospital-core/pull/198) ? Should we just try to produce some reports?
> 2. why are .bak and .old files checked in?
> 3. should both jrxml and jasper files be checked in? IIRC, one is the compiled version of the other, right..?

1. yes, just running reports randomly (or rigourously) to see if libraries are working fine (no matter if no data, anyway you could use create_all_demo.sql)
2. mistake, we can remove them
3. yes, we keep them both, or we can make Jasper to compile them at each application boot, but in another PR maybe

pviotti (approved):
I've tested by just generating a few reports (without demo data) with both Core and Gui built on this branch. From this cursory check, all looks good.

src/main/java/org/isf/serviceprinting/gui/MedicalPrintSelection.java
```diff
@@ -290,5 +298,21 @@ public void actionPerformed(ActionEvent e) {
 	}
 	public void actionPerformed(ActionEvent e) {
 		formatSelected = e.getActionCommand();
+	}
+	
+	public static void main(String[] args) {
```
pviotti, line 303: this main can be removed.

src/main/java/org/isf/serviceprinting/gui/MedicalStockSelection.java
```diff
@@ -501,5 +509,21 @@ public void actionPerformed(ActionEvent e) {
 	public void actionPerformed(ActionEvent e) {
 		formatSelected = e.getActionCommand();
 	}
+	
+	public static void main(String[] args) {
```
pviotti, line 513: this main can be removed too.
(we can figure out another quick way of testing GUI windows - using main methods is not the best solution)

PR #217, opened 2020-09-20
OP-255 Fix PATIENT_PHOTO

mwithi:
OP-255 Solved point 3 of the issue

3. Reports in rpt/ using old field PAT_PHOTO should be updated as well


mwithi:
Superseded by #207

PR #221, opened 2020-10-01
Update startup script - startup.sh

mizzioisf:
- Updated syntax, comments, environment variables
- The script is now independent from file system location (you don't need to be in the OpenHospital installation dir to launch the software)
- Tested on Ubuntu 20.04 x64 - OpenJRE 14 - MariaDB 10.3

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -35,9 +35,9 @@ if [ -z $JAVA_HOME ]; then
   echo "Error - JAVA_HOME not found. Please set it up properly."
   exit
   fi
+  # set JAVA_BIN
   JAVA_BIN=$JAVA_HOME/bin/java
 fi
-
 
 ######## OPENHOSPITAL Configuration
 # OPENHOSPITAL_HOME is the directory where OpenHospital files are located
@@ -48,7 +48,7 @@ if [ -z $OPENHOSPITAL_HOME ]; then
   OPENHOSPITAL_HOME=$PWD
 fi
 
-####### DO NOT EDIT BELOW THIS LINE #######
+######################## DO NOT EDIT BELOW THIS LINE ########################
 
 DIRLIBS=${OPENHOSPITAL_HOME}/bin/*.jar
 for i in ${DIRLIBS}
@@ -100,6 +100,4 @@ esac
 
 ######### OPENHOSPITAL STARTUP
 
-cd $OPENHOSPITAL_HOME
-
-$JAVA_BIN -Dsun.java2d.dpiaware=false -Djava.library.path=${NATIVE_LIB_PATH} -classpath "$OPENHOSPITAL_CLASSPATH" org.isf.menu.gui.Menu "$@"
+cd $OPENHOSPITAL_HOME&& $JAVA_BIN -Dsun.java2d.dpiaware=false -Djava.library.path=${NATIVE_LIB_PATH} -classpath "$OPENHOSPITAL_CLASSPATH" org.isf.menu.gui.Menu "$@"
--- a/startup.sh
+++ b/startup.sh
@@ -23,17 +23,21 @@
 
 ######## Environment check:
 
-# SET JAVA_HOME
+# SET JAVA_BIN or JAVA_HOME
+JAVA_BIN = `which java`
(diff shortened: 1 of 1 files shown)


startup.sh
```diff
@@ -23,21 +23,30 @@
...
-JAVA_EXE=$JAVA_HOME/bin/java
+# SET JAVA_HOME
+#JAVA_HOME=/usr/lib/jvm/java-14-openjdk-amd64/
+
+# check if java home exists
 
 if [ -z $JAVA_HOME ]; then
-  echo "JAVA_HOME not found. Please set it up properly."
-  JAVA_EXE=java
+  echo "Error - JAVA_HOME not found. Please set it up properly."
```
pviotti, line 32: I wonder if this old dance of `JAVA_HOME` env var is still needed. After all we just use it to locate the `java` executable, no? In most (all?) Linux distributions, if you install a JRE, `java` will be in PATH anyway - and `JAVA_HOME` is often not set (it's certainly not set in Arch Linux).  
So maybe this check can be removed?

mizzioisf: You might be right. I guess we can have linux systems with more than one JRE installed (sun and java), so we need to give the user the possibility of choosing which JAVA binary to use. I will try to work on the script to better clarify the choices available and to set-up JAVA_BIN automatically.

mizzioisf (commented):
It should be ok now. We will still have to review the absolute filesystem path in the code, because the script is still directory-dependent: I use a simple trick to cd into the OpenHospital dir where the main application resides.

startup.sh
```diff
@@ -87,4 +98,6 @@ case $ARCH in
 		;;
 esac
 
-$JAVA_EXE -Dsun.java2d.dpiaware=false -Djava.library.path=${NATIVE_LIB_PATH} -classpath "$OPENHOSPITAL_CLASSPATH:$CLASSPATH" org.isf.menu.gui.Menu "$@"
+######### OPENHOSPITAL STARTUP
+
+cd $OPENHOSPITAL_HOME&& $JAVA_BIN -Dsun.java2d.dpiaware=false -Djava.library.path=${NATIVE_LIB_PATH} -classpath "$OPENHOSPITAL_CLASSPATH" org.isf.menu.gui.Menu "$@"
```
pviotti, line 103: nitpick: I'm not a big fan of scripts that change `$PWD`. I suggest using `pushd $OPENHOSPITAL_HOME` and then `popd` after the execution.

mizzioisf: Me neither :-) I commented that it is a bad trick, but there is no other way of having the script working if launched outside of the OPEN_HOSPITAL directory.
That's the reason why I asked how we can strip the references to the absolute paths in the code / properties file.
You can  try the behaviour very simply, just comment out the cd $OPENHOSPITAL_HOME and try something like

cd /tmp
/usr/local/OpenHospital/startup.sh

pviotti: yes, absolute paths in code are definitely something we should change.
in the meantime, my point was more about restoring the PWD the user was in, when finishing the execution - you can do it with `pushd` / `popd`.

pviotti, line 103:
```suggestion
pushd $OPENHOSPITAL_HOME
$JAVA_BIN -Dsun.java2d.dpiaware=false -Djava.library.path=${NATIVE_LIB_PATH} -classpath "$OPENHOSPITAL_CLASSPATH" org.isf.menu.gui.Menu "$@"
popd
```

mizzioisf: I agree with the change but we need to use the Bash shell to have pushd / popd.
Are we willing to use bash instead of the simple (and always available) /bin/sh ?

pviotti: good point. 
I'd say try `cd $OPENHOSPITAL_HOME` and `cd -` instead. 
Or save `$PWD` in a variable and restore it with `cd $myvar`.
Otherwise yes, I'd tend to believe bash or zsh are installed by default in most Linux desktop machines nowadays, so if the above doesn't work, I'd go with push/popd. Thanks.

mizzioisf: Agreed. Modified the script accordingly.

PR #222, opened 2020-10-02
Java version check update

pviotti:
Strictly speaking I'm not sure why that java version check is there: if the version isn't right, it wouldn't be able to execute the bytecode at all, no? If so, should I remove that check?


mwithi:
Nope, the Java6 compliance was for native libraries that are evaluated at runtime and also some lib dependecies that are evaluated at runtime.

Anyway, good to change this, and lost compliance should be documented as per OP-76, OP-77 and OP-228.

PR #226, opened 2020-10-11
OP-95: BillsReport Translated

giuseppesorge:
BillsReport translated in according with OP-95


giuseppesorge:
Closed the pull request. I'll create only one PR for all translated reports.

PR #229, opened 2020-10-22
OP-226 Upgrade to java8

mwithi:
Draft PR to track the *huge* porting to Java8.

This PR is paired with https://github.com/informatici/openhospital-core/pull/232


src/main/java/org/isf/utils/jobjects/LocalDateSupportingJDateChooser.java
```diff
@@ -0,0 +1,57 @@
+package org.isf.utils.jobjects;
```
dbmalkovsky, line 1: File is missing license header; it can be automatically added by running `mvn license:format` in the root directory of the repo.

src/main/java/org/isf/utils/jobjects/LocalDateSupportingJDateChooser.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.toedter.calendar.IDateEditor;
```
dbmalkovsky, line 3: Import order is incorrect; the order is `static` imports, `java`, `javax`, `org`, `com` and then anything else.

PR #233, opened 2020-10-30
OP-95 reports localization

giuseppesorge:
Reports translation


mwithi (approved):
Everything looks fine. Thanks a lot for this huge work!

PR #236, opened 2020-11-07
OP-340 Updating Ward fix

dnlebgwa:
This is to fix OP-340 - Updating Ward more than once leads to error message saying 'The data has been updated by someone else'

### DIFF
--- a/src/main/java/org/isf/ward/gui/WardBrowser.java
+++ b/src/main/java/org/isf/ward/gui/WardBrowser.java
@@ -24,7 +24,6 @@ import org.isf.utils.exception.gui.OHServiceExceptionUtil;
 import org.isf.utils.jobjects.ModalJFrame;
 import org.isf.ward.manager.WardBrowserManager;
 import org.isf.ward.model.Ward;
-import org.springframework.beans.BeansException;
 
 /**
  * This class shows a list of wards.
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/ward/gui/WardBrowser.java
```diff
@@ -48,6 +49,14 @@ public void wardInserted(AWTEvent e) {
...
 	
 	public void wardUpdated(AWTEvent e) {
+		try {
+			ArrayList<Ward> wards = Context.getApplicationContext().getBean(WardBrowserManager.class).getWards(ward);
+			if(wards.size() == 1) {
+				ward = wards.get(0);
+			}
+		} catch (OHServiceException ex) {
+			OHServiceExceptionUtil.showMessages(ex);
+		}
```
mwithi, line 59: It should not be necessary to reload data from the DB, comments under OP-340

src/main/java/org/isf/ward/gui/WardBrowser.java
```diff
@@ -24,6 +24,7 @@
 import org.isf.utils.jobjects.ModalJFrame;
 import org.isf.ward.manager.WardBrowserManager;
 import org.isf.ward.model.Ward;
+import org.springframework.beans.BeansException;
```
dbmalkovsky, line 27: FWIW, this import is not used/needed.

PR #241, opened 2020-11-18
Add JFrog repository to resolve jai-core dependency

dbmalkovsky:
Add Atlassian JFrog repository to resolved missing jai-core jar.   Same fix as in the core repository.


dbmalkovsky:
This is the wrong fix; the fix is really required in the CORE repository; see https://github.com/informatici/openhospital-core/pull/244.

Closing this PR.

PR #242, opened 2020-11-18
Removing repo-spring-io

mwithi:
following Spring notice:
https://spring.io/blog/2020/10/29/notice-of-permissions-changes-to-repo-spring-io-fall-and-winter-2020

Paired with informatici/openhospital-core#248.


mwithi:
Paired with https://github.com/informatici/openhospital-core/pull/248.

PR #246, opened 2020-11-25
Op 262 contextualize weight and height

mwithi:
**Paired with https://github.com/informatici/openhospital-core/pull/254 (need its merge for the checks to pass)**

Found out that the @Transient weight property in Patient entity was used only (1 occurrence) in WardPharmacyNew, while @Transient height property never used.

Going to simplify Patient’s entity and to retrieve his/her weight only when needed.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
@@ -34,6 +34,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.EventListener;
 import java.util.GregorianCalendar;
+import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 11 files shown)


src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
```diff
@@ -124,9 +128,17 @@ public void patientSelected(Patient patient) {
 		jButtonPickPatient.setText(MessageBundle.getMessage("angal.medicalstockwardedit.changepatient")); //$NON-NLS-1$
 		jButtonPickPatient.setToolTipText(MessageBundle.getMessage("angal.medicalstockwardedit.changethepatientassociatedwiththismovement")); //$NON-NLS-1$
 		jButtonTrashPatient.setEnabled(true);
-//		if (patientSelected.getWeight() == 0) {
-//			JOptionPane.showMessageDialog(WardPharmacyNew.this, MessageBundle.getMessage("angal.medicalstockwardedit.theselectedpatienthasnoweightdefined"));
-//		}
+		
+		ExaminationBrowserManager exm = Context.getApplicationContext().getBean(ExaminationBrowserManager.class);
+		try {
+			PatientExamination lastExam = exm.getLastByPatID(patientSelected.getCode());
```
tomek39856, line 134: as we have java8 maybe Optional.ofNullable(exm.getLastByPatID(patientSelected.getCode()))
.map(lastExam -> lastExam.getPex_weight())
.map(weight -> weight.floatValue())
.ifPresent(w -> patientWeight = w);

tomek39856 (changes requested):

src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
mwithi: is it not unncessary complicated? 😅
streams processing is great for iterations, but for a single getter I think it's overcomplicated and decrease a lot readability... map iterations could be even inefficient if compared to previous code, no?

tomek39856: Depends on what you are used to :) I try to not use if (x == null) checks at all and find it more readable. Moreover with map it's 100% null safe also if wieght is null

mwithi: you just reminded me that I would have missed a null check on the `getPex_weight()` return value! Java6 vs Java8: 0 - 1 😄

I was just wondering about the overuse of such pattern... although readability may depend on someone habits, efficiency do not: .ofNullable, .map, .ifPresent are all iterations (they all perform an if-iteration pattern interally), and they are 4 against 2 (if we add the missing `null` check in the previous code).

However, for WIW efficiency at this level, I'm going to apply your suggested solution if you did the same all over https://github.com/informatici/openhospital-gui/pull/220 and OP-226 in general 👍

Thanks a lot!

PR #394, opened 2021-01-08
OP-403 Assign to visits instead of vsrows

dbmalkovsky:
This it a companion fix to the push to Core for OP-403.  See https://github.com/informatici/openhospital-core/pull/288

### DIFF
--- a/src/main/java/org/isf/visits/gui/VisitView.java
+++ b/src/main/java/org/isf/visits/gui/VisitView.java
@@ -162,17 +162,11 @@ public class VisitView extends ModalJFrame {
 	private void loadDataForWard(Ward ward) {
 		try {
 			if (ward != null)
-				vsRows = vstManager.getVisitsWard(ward.getCode());
+				visits = vstManager.getVisitsWard(ward.getCode());
 			else
-				vsRows = vstManager.getVisitsWard(null);
+				visits = vstManager.getVisitsWard(null);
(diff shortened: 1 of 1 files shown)


mwithi (changes requested):
It seems it needs more deep analysis because vsRows would be not initialized and so here a NPE would be rised:
https://github.com/informatici/openhospital-gui/blob/2014404eae03fe776c418366891a349b219a9b3f/src/main/java/org/isf/visits/gui/VisitView.java#L350

dbmalkovsky:
Fixed the NPE by removing `vsRows` definition and update; it was never queried.

mwithi:
> Fixed the NPE by removing `vsRows` definition and update; it was never queried.

Fine. What about the compilation error reported at OP-403 ?

dbmalkovsky:
@mwithi  I'm not seeing it; what exactly is it reporting and what version of the core jar are you using.  Note that the CI build is clean.

mwithi:
> @mwithi I'm not seeing it; what exactly is it reporting and what version of the core jar are you using. Note that the CI build is clean.

It is clean because https://github.com/informatici/openhospital-core/pull/288 is not yet merged. I'm reporting that the two PR supposed to work together probably need a check https://openhospital.atlassian.net/browse/OP-403?focusedCommentId=12976
image

mwithi (approved):

PR #396, opened 2021-01-16
OP-408 Fixing "The data has been updated by someone else”

dnlebgwa:
OH-GUI fix for OP-408

### DIFF
--- a/src/main/java/org/isf/disease/gui/DiseaseEdit.java
+++ b/src/main/java/org/isf/disease/gui/DiseaseEdit.java
@@ -302,14 +302,25 @@ public class DiseaseEdit extends JDialog {
 					disease.setIpdOutInclude(includeIpdOutCheckBox.isSelected());
 					
 					boolean result = false;
+					Disease savedDisease = null;
 					try{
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/disease/gui/DiseaseEdit.java
```diff
@@ -302,14 +302,25 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
 					disease.setIpdOutInclude(includeIpdOutCheckBox.isSelected());
 					
 					boolean result = false;
+					Disease savedDisease = null;
 					try{
 						if (insert) { // inserting
-							result = manager.newDisease(disease);
+							savedDisease = manager.newDisease(disease);
+							if(savedDisease != null) {
```
dbmalkovsky, line 309: This is just a code style comment; the project uses `if (` instead of `if(` [a space between `if` and `(`] as you have added in this PR.  Thanks.

dnlebgwa: @dbmalkovsky I have fixed the code style issue

PR #399, opened 2021-01-26
Chore: update JavaDoc

dbmalkovsky:
Update JavaDoc based on IntelliJ's Analyze JavaDoc option.

Before changes:  8 errors, 184 warnings
After changes: 78 warnings (all are missing the description associated with a method parameter or return value)

Examples of changes:

- Remove empty JavaDoc block comments
- Remove "@return void"
- Remove dangling JavaDoc comments; those are comments that start with "/**" but aren't closed; just add a space so it is "/* *"
- Grammar and spelling


mwithi (approved):
Thanks @dbmalkovsky, very needed!

PR #402, opened 2021-01-27
Chore: remove self assignment of variable

dbmalkovsky:
Found with SpotBugs and I reported as a review comment on [PR 175](https://github.com/informatici/openhospital-gui/pull/175).

The method parameter is assigned to itself and then used only once in the last line of the method.


mwithi (approved):
Thanks!

PR #403, opened 2021-01-29
OP-420 fix null checking

dbmalkovsky:
As described in OP-420, expand the null checks into separate conditionals to stop the possibility of NPEs.

Found with SpotBugs.


mwithi (approved):
LGTM, thanks a lot!

PR #404, opened 2021-01-29
Chore: remove redundant interface declaration

dbmalkovsky:
Remove the redundant interface declaration as `AbstractTreeTableModel` already implements `TreeTableModel `.

Found with Intellij's Analyze plugin looking for redundant code.


mwithi (approved):
Thanks!

PR #405, opened 2021-01-29
OP-251 OP-158 changes not needed since Java8

mwithi:
Trying to revert changes in OP-158 since now Java8 has been adopted for the project and so the underneath EventQueue management is changed.

Gui events deadlocks happening with Java6 should no longer happen (see OP-158 for more details).


mwithi:
Superseded https://github.com/informatici/openhospital-core/pull/354/files

PR #406, opened 2021-01-29
OP-421 remove unneeded contains() method

dbmalkovsky:
Remove unnecessary `contains()` method as described in OP-421.


mwithi (approved):
Thanks!

PR #391, opened 2020-12-23
OP-372 fixed OperationRow language dependence.

gildasdongmo:
The operation row no longer depends on the language, the code is stored and translated on the display.
The core module have been updated for it.

PR #407, opened 2021-01-31
Chore: remove redundant cast where cast is the same as the variable is defined

dbmalkovsky:
I only removed the most **obvious** redundant type casts: where the cast matched the type of the object exactly (either defined that way or returned from a method call).   There are many redundant casts that are more subtle.


dbmalkovsky:
Needs changes committed in https://github.com/informatici/openhospital-gui/pull/391 to compile cleanly

dbmalkovsky:
Rebased and started a new CI job; all is now clean.

mwithi (approved):
Very good. Thanks!

PR #419, opened 2021-02-16
OP-441 remove unwritten field and uncalled method

dbmalkovsky:
See OP-441 for description of logic for removing field and method.


mwithi (changes requested):
It seems that also JLabel chooselabel is never used...

dbmalkovsky:
Removed the field, removed unused import, and use single quotes for single character constant.

mwithi (approved):

PR #420, opened 2021-02-16
OP-442 remove unread fields

dbmalkovsky:
See OP-442 for description of logic for removing the fields.

I also did some minor manual reformatting of the code without any other operational changes.

BTW, why is there a hardcoded FRENCH locale in this class?
````
final DateFormat currentDateFormat = DateFormat.getDateInstance(DateFormat.SHORT, Locale.FRENCH);
````

### DIFF
--- a/src/main/java/org/isf/utils/jobjects/OhTableDrugsModel.java
+++ b/src/main/java/org/isf/utils/jobjects/OhTableDrugsModel.java
@@ -30,6 +30,7 @@ import java.util.Locale;
 import javax.swing.event.TableModelListener;
 import javax.swing.table.TableModel;
 
+import org.isf.generaldata.GeneralData;
 import org.isf.generaldata.MessageBundle;
 import org.isf.medicals.model.Medical;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mwithi** [comment]
> BTW, why is there a hardcoded FRENCH locale in this class?

Dirty code... can we replace with `getDateInstance(DateFormat.SHORT, new Locale(GeneralData.LANGUAGE))` ?

We could also remove all the following, in those method that are actually implemented
```
// TODO Auto-generated method stub
```

Only the `TODO ` keyword for the others that must be there empy for the interface

dbmalkovsky:
Fixed the `Locale` and the `TODO`s.

mwithi:
> Fixed the `Locale` and the `TODO`s.

Sorry I was hasty... I meant to remove the `TODO`s because there's nothing to do...

dbmalkovsky:
`TODO`s removed.

mwithi:
> `TODO`s removed.

Thanks a lot!

PR #425, opened 2021-02-27
Chore: simplify JLabledRequired

dbmalkovsky:
When "fixing" naming issues came across the variable `MARK` that had the style of a constant but wasn't `static final`.  In the process of reviewing the code to see if if fact `MARK` was a constant noticed that the class was called once and only once by `org.isf.supplier.gui.SupplierEdit` when creating a new supplier. 

Removed unnecessary methods that just called super as they weren't used; added some Javadoc, and made some other small improvements. 

Made the changes; rebuilt the project, ran the GUI and saw that the "add new supplier" still functioned properly.


mwithi (approved):
Thanks!

PR #426, opened 2021-02-28
OP-454 non-static non-final upper-case names

dbmalkovsky:
See OP-454.
If a variable is a `constant` make it upper-case with `static final`.
If a variable is not a `constant` make it lower-case with Camel casing.

I also took the liberty of translating some Italian into English.

### DIFF
--- a/src/main/java/org/isf/sms/gui/SmsEdit.java
+++ b/src/main/java/org/isf/sms/gui/SmsEdit.java
@@ -113,7 +113,7 @@ public class SmsEdit extends JDialog implements SelectionListener {
 	}
 	
 	private void initialize() {
-		maxLength = smsManager.getMAX_LENGHT();
+		maxLength = smsManager.getMaxLength();
 	}
 
 	private void initComponents() {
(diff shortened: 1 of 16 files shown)


src/main/java/org/isf/sms/gui/SmsEdit.java
```diff
@@ -113,7 +113,7 @@ public SmsEdit(JFrame owner) {
 	}
 	
 	private void initialize() {
-		MAX_LENGTH = smsManager.getMAX_LENGHT();
+		maxLength = smsManager.getMAX_LENGHT();
```
mwithi, line 116: should be renamed the getter also?

mwithi (changes requested):
Just one small change required.
Thanks a lot!

mwithi (approved):

src/main/java/org/isf/sms/gui/SmsEdit.java
mwithi: Thanks!

PR #427, opened 2021-03-02
OP-457 use camel case names

dbmalkovsky:
See OP-457.

Until the corresponding [PR](https://github.com/informatici/openhospital-core/pull/346) in Core is committed these changes will fail.


mwithi (approved):
Thanks a lot!

PR #429, opened 2021-03-04
Chore: use contains() instead of indexOf()

dbmalkovsky:
Found with IntelliJ Analyze migration aids inspection.

Use `contains()` instead of `indexOf()`; if nothing else it is more descriptive of the purpose of the test.

I got carried away and did some "reformatting" in the process.


mwithi (approved):
Thanks!

PR #431, opened 2021-03-06
OP-462 Unnecessary return statement in void method

dbmalkovsky:
See OP-462

Lots of files but very little change.


mwithi (approved):
Thanks a lot!

PR #432, opened 2021-03-08
OP-463 remove unused collections/variables

dbmalkovsky:
See OP-463

Remove collections/variables that were added to or cleared but never read.

Also a little reformatting/improvements; very minor.


mwithi (approved):
Thanks a lot!

PR #433, opened 2021-03-09
OP-464 use isEmpty()

dbmalkovsky:
See OP-464

### DIFF
--- a/src/main/java/org/isf/accounting/gui/PatientBillEdit.java
+++ b/src/main/java/org/isf/accounting/gui/PatientBillEdit.java
@@ -1182,7 +1182,7 @@ public class PatientBillEdit extends JDialog implements SelectionListener {
 	}
 
 	private boolean hasNewPayments() {
-		return (insert && payItems.size() > 0) || (payItems.size() - payItemsSaved) > 0;
+		return (insert && !payItems.isEmpty()) || (payItems.size() - payItemsSaved) > 0;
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/accounting/gui/PatientBillEdit.java
```diff
@@ -1373,7 +1373,7 @@ public void actionPerformed(ActionEvent e) {
 	private boolean isValidPaymentDate(GregorianCalendar datePay) {
 		GregorianCalendar now = new GregorianCalendar();
 		GregorianCalendar lastPay = new GregorianCalendar();
-	    if (payItems.size() > 0) { 
+	    if (!payItems.isEmpty()) {
 	    	lastPay = payItems.get(payItems.size()-1).getDate();
 		} else {
 			lastPay = billDate;
```
mwithi, line 1379: Strange indentation

dbmalkovsky: I'll fix it; it is most likely a tabs vs. spaces issue.

dbmalkovsky: Reformatted the entire method so it should be consistent.

mwithi: Thanks!

PR #437, opened 2021-03-11
Chore use properties to ease changes

dbmalkovsky:


mwithi (commented):
Less important than on "api" but still much appreciated! Thanks!

PR #447, opened 2021-03-22
OP-339 resolve hard coded paths for configurations files

mwithi:
See OP-339
To be checked with https://github.com/informatici/openhospital-core/pull/372

### DIFF
--- a/src/main/java/org/isf/dicom/gui/DicomGui.java
+++ b/src/main/java/org/isf/dicom/gui/DicomGui.java
@@ -47,7 +47,6 @@ import javax.swing.JSplitPane;
 import javax.swing.LayoutStyle;
 import javax.swing.ScrollPaneConstants;
 
-import org.apache.log4j.PropertyConfigurator;
 import org.isf.admission.gui.PatientFolderBrowser;
 import org.isf.dicom.manager.DicomManagerFactory;
 import org.isf.dicom.manager.SourceFiles;
@@ -515,7 +514,6 @@ public class DicomGui extends JFrame implements WindowListener {
 	 * @param args
 	 */
 	public static void main(String[] args) {
-//		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
 		ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");
 		Context.setApplicationContext(context);
 		GeneralData.getGeneralData();
--- a/src/main/java/org/isf/dicom/gui/DicomGui.java
+++ b/src/main/java/org/isf/dicom/gui/DicomGui.java
@@ -515,7 +515,7 @@ public class DicomGui extends JFrame implements WindowListener {
 	 * @param args
 	 */
 	public static void main(String[] args) {
-		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
+//		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
 		ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");
 		Context.setApplicationContext(context);
 		GeneralData.getGeneralData();
--- a/src/main/java/org/isf/dicom/gui/DicomGui.java
+++ b/src/main/java/org/isf/dicom/gui/DicomGui.java
@@ -516,7 +516,7 @@ public class DicomGui extends JFrame implements WindowListener {
 	public static void main(String[] args) {
 		ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");
 		Context.setApplicationContext(context);
-		GeneralData.getGeneralData();
+		GeneralData.initialize();
 		PatientBrowserManager patManager = Context.getApplicationContext().getBean(PatientBrowserManager.class);
 		Patient patient = new Patient();
 		try {
--- a/src/main/java/org/isf/menu/gui/MainMenu.java
+++ b/src/main/java/org/isf/menu/gui/MainMenu.java
@@ -43,6 +43,7 @@ import javax.swing.JLabel;
 import javax.swing.JPanel;
 
 import org.isf.generaldata.GeneralData;
+import org.isf.generaldata.MessageBundle;
 import org.isf.menu.manager.Context;
 import org.isf.menu.manager.UserBrowsingManager;
 import org.isf.menu.model.User;
@@ -124,6 +125,7 @@ public class MainMenu extends JFrame implements ActionListener, Login.LoginListe
 		myFrame = this;
 
 		GeneralData.getGeneralData();
+		MessageBundle.getBundle();
 		try {
 			singleUser = GeneralData.SINGLEUSER;
 			internalPharmacies = GeneralData.INTERNALPHARMACIES;
--- a/src/main/java/org/isf/menu/gui/MainMenu.java
+++ b/src/main/java/org/isf/menu/gui/MainMenu.java
@@ -124,10 +124,9 @@ public class MainMenu extends JFrame implements ActionListener, Login.LoginListe
 	public MainMenu() {
 		myFrame = this;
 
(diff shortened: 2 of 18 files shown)


src/main/java/org/isf/dicom/gui/DicomGui.java
```diff
@@ -515,7 +515,7 @@ public void detail() {
 	 * @param args
 	 */
 	public static void main(String[] args) {
-		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
+//		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
```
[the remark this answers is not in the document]
mwithi: yep! it was just temporary

src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -124,6 +125,7 @@ public MainMenu() {
 		myFrame = this;
 
 		GeneralData.getGeneralData();
+		MessageBundle.getBundle();
```
dbmalkovsky, line 128: What is the purpose of this statement; it returns a `ResourceBundle` but it already has to exist and this statement is not even saving the returned value?

I will also point out that `GeneralData.getGeneralData()` returns a singleton `GeneralData` that is never assigned in the 11 occurrences in OH.    And then the fact that `getGeneralData` is responsible for initializing the ResourceBundle is also less than obvious.

If the methods are not going to do anything with the returned value then instead of `getGeneralData()` it should be something like `initGeneralData()`.   And instead of burying the ResourceBundle init maybe the init routine should be called directly.

mwithi: true, for MessageBundle it can be initialized with the `initialize()` method; anyway, both cases "getter" is an heritage of previous architecture. According to GOF, we could rename all getters of singleton classes into `getInstance()`, in this way whereas the returned object is used or not the call is more explicit, wdyt?

emecas (commented):
Compilation problem:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  6.730 s
[INFO] Finished at: 2021-03-23T07:44:16-05:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project OH-gui: Could not resolve dependencies for project org.isf:OH-gui:jar:1.11: Could not find artifact org.isf:OH-core:jar:1.11 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
[ERROR]
```

dbmalkovsky:
I think you need to do a `mvn install` locally on the latest core project and then the compile will work.

src/main/java/org/isf/menu/gui/MainMenu.java
dbmalkovsky: Yes it makes it more meaningful but just barely IMHO as there is still no *getting*.   All the method is doing is initializing a global object that is accessed through `public static` variables like `GeneralData.OPDEXTENDED`; as it now written there is no reason ever to get a handle to the instance.

emecas:
> I think you need to do a `mvn install` locally on the latest core project and then the compile will work.

you right , wrong order, I did not even noticed dependency missed was CORE thx

src/main/java/org/isf/dicom/gui/DicomGui.java
```diff
@@ -515,7 +514,6 @@ public void detail() {
 	 * @param args
 	 */
 	public static void main(String[] args) {
-		PropertyConfigurator.configure(new File("./rsc/log4j.properties").getAbsolutePath());
 		ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");
```
emecas, line 517: here is a hard coded file name, is this fine ?

src/main/java/org/isf/menu/gui/Menu.java
```diff
@@ -72,7 +71,6 @@ public static void checkJavaVersion() {
 	}
 
 	public static void main(String[] args) {
```
emecas, line 73: Just from my ignorance about GUI project , why multiple `main` methods ? how are these used?

dbmalkovsky: My reading is that the multiple `main`s were used for testing purposes and as the testing matures they can be removed.

src/main/java/org/isf/dicom/gui/DicomGui.java
mwithi: you mean the removed line?

src/main/java/org/isf/menu/gui/Menu.java
mwithi: well, "testing gui" is not only JUnit, but also "see" the resulting JFrame/JDialog and check the graphic layout... having the main with some mock data can speed up the thing

src/main/java/org/isf/menu/gui/MainMenu.java
mwithi: Fair enough, but this let me think also another thing: having public static variables allow any piece of code to access them even if not initialized, without knowing if they host the intended values or not, and even changing them afterward; this can be fine to programmatically change some behviors but it can lead to security problem if we see at SINGLEUSER variable in GeneralData (which should be final once intialized...). All the changes in this PR were meant to accomodate actual architecture, but maybe we have to rethink it...

src/main/java/org/isf/menu/gui/Menu.java
emecas: OK, makes sense, if that is the point keep in mind that  is susceptible to automation too. I think lifetime for GUI project could be short at this point or maybe not, but I used [this](https://joel-costigliola.github.io/assertj/assertj-swing.html) for that kind of testing  some time ago!!!

src/main/java/org/isf/dicom/gui/DicomGui.java
emecas: No, I meant the line that is going to stay in the code, and any other line hardcoding file names, since that is the goal or the PR :)  

 517  ->  `ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");`

src/main/java/org/isf/menu/gui/Menu.java
mwithi: yes, but as you say the GUI project lifetime could make the effort worthless, thanks anyway to pointed it out!

mwithi: Note that OP-226 effort are headed to isolate and test the GUI logic only

src/main/java/org/isf/dicom/gui/DicomGui.java
mwithi: yeah, filename is ok, the issue is about hardcoded "paths".
Anyway, in this PR we are only caring about configuration files, other paths that could be resolved are the ones for icons and images.

src/main/java/org/isf/menu/gui/MainMenu.java
mwithi: To be more clear I've pushed an update: 1) both initialize() and getter() 2) private final SINGLEUSER with getter;
Could it be a valid intermediate solution for the above?

dbmalkovsky: Looks better thanks; and +1 as an intermediate solution.

mwithi: Thanks to you also!

mizzioisf (approved):
After a large number of tests:

- logging: log4 jconfiguration is now path indipendent - OK
- images, icons and splashscreen don't work by inclusion in the CLASSPATH, but I think this is not a big problem: I will keep the "cd $OH_DIR" workaround in the script, in order to launch OH-gui always from the correct directory.


I would proceed with the merge !

PR #448, opened 2021-03-23
Chore performance

dbmalkovsky:
Original change was to "unbox" already "boxed" item:  `new Integer(origin.getSupId())` as `origin.getSupId()` returns an `Integer` object thus no need to create another `Integer` object (see line 1250).

 Decided to change strings of length one to character objects as there may be an performance benefit in some cases.

Found with IntelliJ's Analyze plugin.


mwithi (approved):
Thanks!

PR #449, opened 2021-03-23
OP-474 update version checking

dbmalkovsky:
See OP-474 as the reasoning behind the new testing method.


mwithi (approved):
LGTM. Thanks a lot!

Anyway not possible to test since with java less than 1.8 simply would not compile.

PR #450, opened 2021-03-24
OP-475 remove unnecessary unboxing

dbmalkovsky:
See OP-475

Got carried away and did a little reformatting to make things consistent.


mwithi (approved):
Thanks a lot!

PR #451, opened 2021-03-25
OP-476 remove unnecessary boxing

dbmalkovsky:
See OP-476

In the process removed a parameter to a method that was no longer used; `protected BigDecimal askCost()` in WardPharmacyRectify.


mwithi (approved):
All good. Thanks!

PR #453, opened 2021-03-30
OP-482 set date to end of day for comparison

dbmalkovsky:
See OP-482

Made the change and tested it with the example given in the JIRA issue

### DIFF
--- a/src/main/java/org/isf/utils/jobjects/DateTextField.java
+++ b/src/main/java/org/isf/utils/jobjects/DateTextField.java
@@ -179,6 +179,8 @@ public class DateTextField extends JPanel{
 		date.set(Calendar.DAY_OF_MONTH,getDay());
 		date.set(Calendar.MONTH,getMonth()-1);
 		date.set(Calendar.YEAR,getYear());
+		// This is a temporary solution so the comparison is only on the "date" part of "date and time"
+		// See comments in OP-482: https://openhospital.atlassian.net/browse/OP-482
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/utils/jobjects/DateTextField.java
```diff
@@ -175,9 +176,12 @@ public GregorianCalendar getCompleteDate(){
...
 		}
-		date.set(GregorianCalendar.DAY_OF_MONTH,getDay());
-		date.set(GregorianCalendar.MONTH,getMonth()-1);
-		date.set(GregorianCalendar.YEAR,getYear());
+		date.set(Calendar.DAY_OF_MONTH,getDay());
+		date.set(Calendar.MONTH,getMonth()-1);
+		date.set(Calendar.YEAR,getYear());
+		date.set(Calendar.HOUR_OF_DAY, 23);
+		date.set(Calendar.MINUTE, 59);
+		date.set(Calendar.SECOND, 59);
```
mwithi, line 184: As said in comments on OP-482, this is accettable as temporary solution, could we add a comment for this? Thanks!

mwithi (approved):
LTGM, only one comment required

src/main/java/org/isf/utils/jobjects/DateTextField.java
dbmalkovsky: @mwithi FWIW,  I added the comment.

mwithi (approved):
Thanks a lot!

PR #479, opened 2021-04-28
OP-497 Standardized on slf4j.Logger

dbmalkovsky:
See OP-497.

Note the spacing appears to be inconsistent and that is because these files use multiple formatting techniques.  A reformat of the entire file to match standards is the real answer.

### DIFF
--- a/src/main/java/org/isf/lab/gui/LabNew.java
+++ b/src/main/java/org/isf/lab/gui/LabNew.java
@@ -35,8 +35,6 @@ import java.util.ArrayList;
 import java.util.EventListener;
 import java.util.GregorianCalendar;
 import java.util.Locale;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 import javax.swing.BorderFactory;
 import javax.swing.BoxLayout;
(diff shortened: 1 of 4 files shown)


mwithi (approved):
LGTM. Anyway, I know that `LOGGER` is `static final` but in the rest of the application we declared lowercase, should we keep this convention?

dbmalkovsky:
My convention for `private static final` is to make the variable all upper case.   And I find that is general true for the GUI project.   I don't know why most of the `logger` variables are not declared `static`?   IMHO, they all should be and therefore uppercase.

mwithi:
Actually I don't see the need to make logger fields static, since their purpose is just to make them available to the specific instance of a class, or am I wrong? As you mentioned GUI I've realized we have similar cases not using slf4j.Logger... and some are defined only static, others only final, others both.

If you agree there's no need of have them static, the solution here is an easy thing: let's make them final (only) and so lowercase, wdyt?

dbmalkovsky:
I disagree and so do static analyzers like Sonar; see https://rules.sonarsource.com/java/tag/convention/RSPEC-1312

mwithi:
> I disagree and so do static analyzers like Sonar; see https://rules.sonarsource.com/java/tag/convention/RSPEC-1312

Fair enough (".. to prevent creating a logger along with each object."). 

Anyway I personally don't like the uppercase because we are not define a constant (as used for primitive types) but just saying that the variable is pointing to a class that should be shared and not modified, but we are talking about conventions... I personally prefer it still looking "nice" when calling a method upon it.

The project seems mostly using lowercase for such case, that's why my comment, but I think this is a really trivial subject at this point and we can revise/address in another task.

Sure another, more important, task will be changing them to `static final` as suggested.

Thanks a lot!

PR #483, opened 2021-05-04
OP-503 new GUI helper classes

dbmalkovsky:
See OP-503.

FWIW, I have been using both classes for many days as part of the GUI changes.


mwithi (approved):
Looks nice, thanks!

PR #485, opened 2021-05-05
OP-504 GUI changes for menus, Login, and User Browser

dbmalkovsky:
See OP-504 and attached document that shows the before and after of the changes.

[OP-504.pdf](https://github.com/informatici/openhospital-gui/files/6429380/OP-504.pdf)

Note: there is a small CORE [PR](https://github.com/informatici/openhospital-core/pull/394) component to the changes.


src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -341,12 +334,8 @@ private void launchApp(String itemMenuCode) {
...
 							}
 						}
-					} catch (InstantiationException ie) {
-						ie.printStackTrace();
-					} catch (IllegalAccessException iae) {
-						iae.printStackTrace();
-					} catch (ClassNotFoundException cnfe) {
-						cnfe.printStackTrace();
+					} catch (InstantiationException | ClassNotFoundException | IllegalAccessException ie) {
+						LOGGER.error("", ie);
```
mwithi, line 338: Shall we log something here?

src/main/java/org/isf/menu/gui/UserBrowsing.java
```diff
@@ -121,250 +86,239 @@ public void userUpdated(AWTEvent e) {
...
 	private ArrayList<User> pUser;
 	private String[] pColumns = {
-			MessageBundle.getMessage("angal.menu.userm"), 
-			MessageBundle.getMessage("angal.menu.groupm"), 
-			MessageBundle.getMessage("angal.menu.descm") };
-	private int[] pColumnWidth = {70, 70, 150 };
+			MessageBundle.getMessage("angal.userbrowser.user.txt"),
+			MessageBundle.getMessage("angal.userbrowser.group.txt"),
+			MessageBundle.getMessage("angal.userbrowser.description.txt") };
+	private int[] pColumnWidth = { 70, 70, 150 };
```
mwithi, line 92: We were using `angal.*.<key>m` format as reminder to use UPPERCASE translations for tables' columns.
Shall we get rid of them, having only one translation, and apply `.toUpperCase()` in those cases?

src/main/java/org/isf/menu/gui/UserBrowsing.java
```diff
@@ -121,250 +86,239 @@ public void userUpdated(AWTEvent e) {
...
-                    }
-                }
+
+		JButton buttonDelete = new JButton(MessageBundle.getMessage("angal.common.delete.btn"));
+		buttonDelete.setMnemonic(MessageBundle.getMnemonic("angal.common.delete.btn.key"));
+		buttonDelete.addActionListener(event -> {
+			if (table.getSelectedRow() < 0) {
+				MessageDialog.error(null, "angal.common.select.row.msg");
+			} else {
+				User m = (User) model.getValueAt(table.getSelectedRow(), -1);
```
mwithi, line 258: Shall we give more meaningful name to the variable `m`?

src/main/java/org/isf/menu/gui/UserBrowsing.java
```diff
@@ -121,250 +86,239 @@ public void userUpdated(AWTEvent e) {
...
 		Toolkit kit = Toolkit.getDefaultToolkit();
 		Dimension screensize = kit.getScreenSize();
-		pfrmWidth = screensize.width / 2 + 100;
+		pfrmWidth = (screensize.width / 2) + 100;
 		pfrmHeight = screensize.height / 2;
-		setBounds(screensize.width / 4, screensize.height / 4, pfrmWidth,
-				pfrmHeight);
-		
+		setBounds(screensize.width / 4, screensize.height / 4, pfrmWidth, pfrmHeight);
+
```
mwithi, line 114: We should be able to get rid of these sizing still used in many classes and try `pack(); setLocationRelativeTo(null);` combination

mwithi (commented):
Nice brushing! Just a few comments

src/main/java/org/isf/menu/gui/UserBrowsing.java
dbmalkovsky: In my review it appears that most, but not all, table columns were upper cased.  But if that is the project convention then I will use `.toUpperCase()` when filling in column headers.

bundle/language_en.properties
```diff
@@ -1,4 +1,74 @@
...
+angal.common.ok.btn                                                              = OK
+angal.common.ok.btn.key                                                          = O
+angal.common.submit.btn                                                          = Submit
+angal.common.submit.btn.key                                                      = S
+
+# Other common messages and text
+angal.common.select.row.msg                                                      = Please select a row.
+angal.common.data.not.saved.msg                                                  = The data could not be saved.
+
+angal.common.userid.label                                                        = Userid:
```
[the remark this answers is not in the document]
dbmalkovsky: I will change to "User".

As for colons for labels I like them and most "standards" use them too.   Look at a Find or Replace dialog for example.

But I will do whatever the project dictates.

src/main/java/org/isf/menu/gui/MainMenu.java
dbmalkovsky: Added message instead of stack trace.

src/main/java/org/isf/menu/gui/UserBrowsing.java
dbmalkovsky: done

src/main/java/org/isf/menu/gui/UserBrowsing.java
dbmalkovsky: I will look into this later if that is OK?

src/main/java/org/isf/menu/gui/UserBrowsing.java
mwithi: Thanks!

bundle/language_en.properties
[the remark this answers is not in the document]
mwithi: > As for colons for labels I like them and most "standards" use them too. Look at a Find or Replace dialog for example.

Yes, most modern styles don't even put a label, like Thunderbird:
image

or Chrome:
image

But I'm thinking about translators and the way the [Transifex Translation Memory](https://docs.transifex.com/translation-memory/tm-introduction) (TM) will treat them because I'm not sure yet if it cares about punctuation or not.

So the point is: in cases they are present, are the colons "part of" the label OR are they added programmatically to the localized message?

src/main/java/org/isf/menu/gui/MainMenu.java
mwithi: Thanks!

bundle/language_en.properties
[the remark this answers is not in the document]
dbmalkovsky: I can find counter examples all over; this one is from Outlook on Windows:
image

Linux Text Edit
image

Linux Default Printer dialog
image

My default behavior and previous projects is to require the colons as part of the label but if the only way I'll get them is to add them programmatically then so be it 😄

src/main/java/org/isf/menu/gui/UserBrowsing.java
dbmalkovsky: Used your suggestion; it works although the dialog is now a little smaller in size.

mwithi: Yep, it is the effect of packing, the layout manager expects we play with padding and margins, if we like

mwithi (approved):
LGTM. Created OP-507 for the doc!

PR #490, opened 2021-05-12
OP-431 NPE and Validation error on Exam selection and saving

GraninNikita:
There is no need to call jTableExams.updateUI when updating the selected laboratory material value. 
Commenting out the updateUI call resolves the NPE while changing the selection in table.

### DIFF
--- a/src/main/java/org/isf/lab/gui/LabNew.java
+++ b/src/main/java/org/isf/lab/gui/LabNew.java
@@ -57,8 +57,6 @@ import javax.swing.ListSelectionModel;
 import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
 import javax.swing.event.EventListenerList;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/lab/gui/LabNew.java
```diff
@@ -751,20 +751,20 @@ private JTable getJTableExams() {
 			listSelectionModel.addListSelectionListener(new ListSelectionListener() {
 				
 				public void valueChanged(ListSelectionEvent e) {
-					// Check that mouse has been released.
```
dbmalkovsky, line 754: These look like just formatting changes that actually makes things out of line; see image:
image

If you really want to change this then I'd suggest using a lambda:
image

mwithi (approved):
It worked! Thanks a lot guys!

PR #510, opened 2021-05-21
Chore: fix wrong bundle key value

dbmalkovsky:
Notice two keys for the MessageBundle were specified incorrectly during testing.


mwithi (approved):
Alright! Thanks!

PR #513, opened 2021-05-24
Chore: remove unused and incorrect method

dbmalkovsky:
The method `visColums` is wrong (trying to find a bundle key for a formatted date) and is not even used so it should be removed.


mwithi (approved):
Thanks!

PR #516, opened 2021-05-24
Chore: reformat large sections of code

dbmalkovsky:
This file had various and different formatting styles.  It appears that some of which was caused by removing blocks of code with comments (now removed).

The only significant code changes were adding the ".msg" suffix to two property bundle keys to match the most recent push of language_en.properties.

### DIFF
--- a/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
+++ b/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
@@ -306,26 +306,6 @@ public class AdmissionBrowser extends ModalJFrame {
 	
 	private JPanel diseaseOut3Panel;
 	
-	//private JComboBox operationBox = null;
-
-	//private JRadioButton operationResultRadioP = null;
-
-	//private JRadioButton operationResultRadioN = null;
-
-	//private JRadioButton operationResultRadioU = null;
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/admission/gui/AdmissionBrowser.java
```diff
@@ -1741,422 +1700,376 @@ private JButton getSaveButton() {
...
+	    Date beginDate;
+	    Date endDate;
+	    if (admission.getAdmDate() != null)
+		    beginDate = admission.getAdmDate().getTime();
+	    else
+		    beginDate = null;
+	    if (admission.getDisDate() != null)
+		    endDate = admission.getDisDate().getTime();
+	    else
+		    endDate = null;
```
mwithi, line 2045: code block missing?

dbmalkovsky: Yeah; the reformatting doesn't automatically add them; I will do that the next time through the file.

dbmalkovsky: It was easy to do now so I did it.

mwithi: Alright, thanks!

PR #517, opened 2021-05-25
OP-517 Changes for Admission/Patient main menu item

dbmalkovsky:
See OP-517.

Also added logging to 2 classes instead of printing a stack trace to standard out and removed some commented out code that appears from the git commits to be there from the dawn of time.


mwithi (approved):
LGTM. Thanks!

PR #518, opened 2021-05-26
OP-518 Use MessageDialog.error/warn/info() methods

dbmalkovsky:
See OP-518.

### DIFF
--- a/src/main/java/org/isf/dlvrtype/gui/DeliveryTypeBrowserEdit.java
+++ b/src/main/java/org/isf/dlvrtype/gui/DeliveryTypeBrowserEdit.java
@@ -42,6 +42,7 @@ import org.isf.generaldata.MessageBundle;
 import org.isf.menu.manager.Context;
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.model.OHExceptionMessage;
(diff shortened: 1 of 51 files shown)


src/main/java/org/isf/dlvrtype/gui/DeliveryTypeBrowserEdit.java
```diff
@@ -215,7 +216,9 @@ public void actionPerformed(java.awt.event.ActionEvent e) {
                             if (result) {
                                 fireDeliveryInserted();
                             }
-                            if (!result) JOptionPane.showMessageDialog(null, MessageBundle.getMessage("angal.dlvrtype.thdatacouldnotbesaved"));
+                            if (!result) {
+                                MessageDialog.error(null, "angal.dlvrtype.thdatacouldnotbesaved");
```
mwithi, line 220: reminder: we could move this bundle to `common` key

dbmalkovsky: Yes; there are a few others like that too.  In this pass I didn't want to touch any of the keys but I will do that soon.

mwithi (approved):
LGTM. Thanks a lot!

PR #520, opened 2021-05-27
Chore: improvements to MessageDialog class

dbmalkovsky:
- Make all dialolgs accept both `getMessage()` and `formatMessage()` type messages.
- Add method to display all exception messages in the same format and reduce duplicate code.


mwithi (approved):
LGTM. Thanks!

PR #522, opened 2021-05-28
OP-521 Standardize window/dialog titles

dbmalkovsky:
See OP-521.

Changes per discussion in Jira issue.


mwithi (approved):
LGTM. Thanks!

PR #523, opened 2021-05-28
Chore: remove System.out.println

dbmalkovsky:
Where there was a need to generate output use Logger instead.    When it was just commented out code it was removed.


mwithi (approved):
Approved! Thanks!

PR #524, opened 2021-05-29
OP-522 remove unused/deprecated classes

dbmalkovsky:
See OP-522.

Searched all of api, core, and gui repositories and found no references.


mwithi (approved):
Thanks!

PR #541, opened 2021-05-31
OP-529 fix table case name in jrxml file

dbmalkovsky:
See OP-529.

Someone with JasperReports installed will need to generate the jasper file equivalent.


mwithi (approved):
Thanks!

PR #544, opened 2021-06-03
OP-537 standardize table column headings

dbmalkovsky:
See OP-537.

### DIFF
--- a/src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
+++ b/src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
@@ -162,10 +162,10 @@ public class AdmittedPatientBrowser extends ModalJFrame implements
 			MessageBundle.getMessage("angal.common.name.txt").toUpperCase(),
 			MessageBundle.getMessage("angal.common.age.txt").toUpperCase(),
 			MessageBundle.getMessage("angal.common.sex.txt").toUpperCase(),
-			MessageBundle.getMessage("angal.admission.cityaddresstelephonenote.txt").toUpperCase(),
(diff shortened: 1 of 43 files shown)


src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
```diff
@@ -162,10 +162,10 @@
 			MessageBundle.getMessage("angal.common.name.txt").toUpperCase(),
 			MessageBundle.getMessage("angal.common.age.txt").toUpperCase(),
 			MessageBundle.getMessage("angal.common.sex.txt").toUpperCase(),
-			MessageBundle.getMessage("angal.admission.cityaddresstelephonenote.txt").toUpperCase(),
-			MessageBundle.getMessage("angal.admission.ward.txt").toUpperCase()
+			MessageBundle.getMessage("angal.admission.cityaddresstelephonenote.col").toUpperCase(),
+			MessageBundle.getMessage("angal.common.ward.txt").toUpperCase()
```
mwithi, line 166: in which case we should have .col or .txt?

mwithi (approved):
LGTM (just one question)

src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
dbmalkovsky: I am using ".col" only when it is a table column heading; if it is both used as a column heading and it other places like dialog labels, or just text then I use ".txt".

mwithi (approved):
Approved! Thanks a lot!

PR #545, opened 2021-06-04
Chore: fix bad quote location

dbmalkovsky:
When checking for strings that should be in a resource bundle, found this misplaced quote.


mwithi (approved):
Thanks!

PR #551, opened 2021-06-07
Chore: fix showExceptions()

dbmalkovsky:
The `showExceptions()` method was looking up the message text in the resource bundle even though the exception processing already did it.


mwithi (approved):
LGTM. Thanks!

PR #552, opened 2021-06-07
Chore: fix the comparison

dbmalkovsky:
Comparing to zero means that the title always read, Loading multiple images: ", even when there was a single image.


mwithi (approved):
Thanks!

PR #557, opened 2021-06-10
OP-523 Part 2

dbmalkovsky:
See OP-523  Part 2


src/main/java/org/isf/dicom/gui/ShowPreLoadDialog.java
```diff
@@ -52,6 +52,7 @@
 import org.isf.utils.exception.OHServiceException;
 import org.isf.utils.exception.gui.OHServiceExceptionUtil;
 import org.isf.utils.jobjects.VoLimitedTextField;
+import org.jivesoftware.smackx.workgroup.ext.macros.MacroGroup;
```
mwithi, line 55: is this import really needed?

mwithi (approved):
Wonderful! Thanks!

src/main/java/org/isf/dicom/gui/ShowPreLoadDialog.java
dbmalkovsky: No it isn't; good catch

PR #558, opened 2021-06-10
OP-544 fix some of the JRadioButton shortcut keys

dbmalkovsky:
This fixes `OperationEdit` and `PatientInsert` and only starts to touch on` PatientInsertExtended` which still has duplicated shortcut key issues.

See OP-544.

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -386,6 +386,10 @@ angal.medicalstockward.chooselot.btn                                            
 angal.medicalstockward.chooselot.btn.key                                         = C
 angal.opd.examination.btn                                                        = Examination
 angal.opd.examination.btn.key                                                    = E
+angal.operation.major.btn                                                        = Major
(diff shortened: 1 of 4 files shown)


mwithi (approved):
LGTM. Thanks!

mwithi:
I would wait to merge OP-523 before to proceed

dbmalkovsky:
I am beginning to think that JRadioButtons shouldn't have shortcut keys at all; it is just too confusing especially in the large more complex dialogs.  What do you think?

mwithi:
> I am beginning to think that JRadioButtons shouldn't have shortcut keys at all; it is just too confusing especially in the large more complex dialogs. What do you think?

I agree, let's have the shourcut for JButton(s) only

dbmalkovsky:
I'm going to close this PR and replace it with a new one that removes the shortcuts for JRadioButtons.

PR #559, opened 2021-06-11
OP-544 remove JRadioButton shortcut keys

dbmalkovsky:
See OP-544

Also added ".btn" suffix to the various JRadioButton key values.

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -23,6 +23,7 @@ angal.common.note.title                                                         
 ########################################
 # Common button text and short cut keys
 ########################################
+angal.common.all.btn                                                             = All
 angal.common.cancel.btn                                                          = Cancel
 angal.common.cancel.btn.key                                                      = C
 angal.common.close.btn                                                           = Close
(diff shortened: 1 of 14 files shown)


bundle/language_en.properties
```diff
@@ -321,6 +323,10 @@ angal.billbrowser.today.btn
 angal.billbrowser.today.btn.key                                                  = T
 angal.groupsbrowser.groupmenu.btn                                                = GroupMenu
 angal.groupsbrowser.groupmenu.btn.key                                            = M
+angal.lab.n.btn                                                                  = N
+angal.lab.p.btn                                                                  = P
```
mwithi, line 327: ok, these are "negative" and "positive" abbeviations

bundle/language_en.properties
```diff
@@ -321,6 +323,10 @@ angal.billbrowser.today.btn
...
+angal.labnew.ip.btn                                                              = IP
+angal.labnew.opd.btn                                                             = OPT
```
mwithi, line 329: Fix value to "OPD"

mwithi (changes requested):
little changes

bundle/language_en.properties
dbmalkovsky: I'll fix it but the original value in the bundle was "OPT"

bundle/language_en.properties
dbmalkovsky: Done

bundle/language_en.properties
mwithi: we just had chance to fix it 😉 thanks!

PR #561, opened 2021-06-11
Chore: miscellaneous changes

dbmalkovsky:
The biggest changes related to resource bundles are:

- More `MessageDialog` usage and corresponding key/value updates
- The last browser tables converted to newer format

Did some easy lambdas and `@Override` code changes

Deleted `OHExceptionUtil` as it isn't used and `MessasgeDialog` handles that case and more


mwithi (approved):
LGTM. Thanks!

PR #563, opened 2021-06-13
OP-545 fix what looks like an infinite loop

dbmalkovsky:
What started as removing a seemingly infinite loop as described in OP-545 turned out to be more involved as the method return was never used or set and the confirmation dialog was really only a message dialog.   It appears that a merge of various PRs didn't go quite right.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
@@ -576,9 +576,8 @@ public class WardPharmacy extends ModalJFrame implements
 		}
 	}
 
-	private MedicalWard showLotDetail(ArrayList<MedicalWard> drug, String me) {
+	private void showLotDetail(ArrayList<MedicalWard> drug, String me) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
```diff
@@ -576,33 +576,28 @@ public Component getTableCellRendererComponent(JTable table, Object value, boole
 		}
 	}
 
-	private MedicalWard showLotDetail(ArrayList<MedicalWard> drug, String me) {
+	private void showLotDetail(ArrayList<MedicalWard> drug, String me) {
 		ArrayList<MedicalWard> dr = new ArrayList<>();
```
mwithi, line 580: maybe for sake of readibility we can change `dr` to `medicalWardList`

mwithi (approved):
LGTM. Thanks!

src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
dbmalkovsky: done

PR #575, opened 2021-06-17
OP-452 WIP resolved all the missing keys found in jrxml files

dbmalkovsky:
I made my best guess as to how fix the 8 issues outlined in OP-452.

I updated the jrxml file but **not** the jasper file.

I updated the base properties file but **not** en_properties files nor any other language.

Marking this as a work in progress so those more knowledgeable can review and comment (and create the required jasper files).


mwithi:
Add required changed and merged with https://github.com/informatici/openhospital-gui/commit/84c71789227b21aed6876754a419f4dbc512c642

PR #579, opened 2021-06-20
OP-542 Italian resource bundle update

dbmalkovsky:
See OP-542.   Attempt at updating Italian resource bundle file.


mwithi (approved):
LGTM. Thanks!

PR #584, opened 2021-06-22
OP-542 French resource bundle update

dbmalkovsky:
See OP-542.  This is an update for French which primarily includes missing keys and some some consistency changes.


mwithi (approved):
LGTM. Thanks!

PR #585, opened 2021-06-22
OP-542 Merge Swahili resource bundle with new keys

dbmalkovsky:


bundle/language_sw.properties
```diff
@@ -1,1480 +1,1493 @@
...
+angal.therapy.checkavailability.btn.key                                          = C
+angal.therapy.addtherapy.btn                                                     = Ongeza Tiba
+angal.therapy.addtherapy.btn.key                                                 = A
+angal.therapy.edittherapy.btn                                                    = Hariri Tiba
+angal.therapy.edittherapy.btn.key                                                = E
+angal.therapy.removetherapy.btn                                                  = Ondoa Tiba
+angal.therapy.removetherapy.btn.key                                              = R
+angal.patientphoto.file.btn                                                      = File
+angal.patientphoto.file.btn.key                                                  = F
+angal.patientphoto.newphoto.btn                                                  = New Photo
```
mwithi, line 448: Picha Mpya

bundle/language_sw.properties
```diff
@@ -1,1480 +1,1493 @@
...
+angal.lab.negativeabbr.btn                                                       = N
+angal.lab.positiveabbr.btn                                                       = P
+angal.labnew.ip.btn                                                              = IP
+angal.labnew.opd.btn                                                             = OPD
+angal.newbill.findpatient.btn                                                    = Find Patient
+angal.newbill.findpatient.btn.key                                                = P
+angal.newbill.changepatient.btn                                                  = Badilisha Mgonjwa
+angal.newbill.changepatient.btn.key                                              = H
+angal.newbill.givechange.btn                                                     = Kupa chenji
+angal.newbill.givechange.btn.key                                                 = G
```
mwithi, line 335: I think the btn.key should at least present in the btn label

bundle/language_sw.properties
```diff
@@ -1,1480 +1,1493 @@
...
+angal.groupsbrowser.title                                                        = Groups Browser
+angal.groupsbrowser.newgroup.title                                               = New Group
+angal.groupsbrowser.editgroup.title                                              = Edit Group
+angal.groupsbrowser.menuitembrowser.title                                        = Menu Item Browser
+angal.malnutrition.malnutritionbrowser.title                                     = Kupitiapitia utapiaji mlo
+angal.malnutrition.newmalnutrition.title                                         = New Malnutrition
+angal.malnutrition.editmalnutrition.title                                        = Edit Malnutrition
+angal.medicals.newmedical.title                                                  = New Medical
+angal.medicals.editmedical.title                                                 = Edit Medical
+angal.medicals.pharmaceuticalbrowser.title                                       = Pharmaceutical Browser
```
mwithi, line 211: If the key is missing, OH should log the missing key and should fallback to English (we have still to understand if and how to report it also in the GUI) but in such case the key is not missing and the translation is same as English, so OH will not report anything.

Is this the intended behavior?

mwithi (approved):
It can be ok as for now, not sure what to do with "same as English" temporary translations

bundle/language_sw.properties
dbmalkovsky: Done.

bundle/language_sw.properties
dbmalkovsky: All the files have every key defined; if there isn't a translation then the English version of the string is used in the file.  This means every key should be found in every file.  I was hoping this would make translation easier and I noticed that some of the files when I started were a mix of languages.

mwithi: Thanks @dbmalkovsky, actually it is not a must that every file has all the keys, definitely it could be the signal that a translation is missing... and it is also what it is done by Transifex

I merged #579, #583 and #584 just to see if Transifex also checks if the translation is different or not from the default (English)

PR #586, opened 2021-06-22
OP-542 Merge Arabic resource bundle with new keys

dbmalkovsky:


bundle/language_ar.properties
```diff
@@ -1,1505 +1,1493 @@
...
+angal.common.select.btn                                                          = Select
+angal.common.select.btn.key                                                      = S
+angal.common.update.btn                                                          = Update
+angal.common.update.btn.key                                                      = U
+
+###########################
+# Common messages and text
+###########################
+angal.common.pleaseselectarow.msg                                                = يرجى اختيار صف
+angal.common.datacouldnotbesaved.msg                                             = The data could not be saved.
```
mwithi, line 70: Same doubts as in https://github.com/informatici/openhospital-gui/pull/585#discussion_r658070144

PR #587, opened 2021-06-23
OP-542 Chinese resource bundle update

dbmalkovsky:


bundle/language_zh_CN.properties
```diff
@@ -1,1426 +1,1493 @@
...
+angal.common.thecodeistoolongmax1char.msg                                        = 代码太长（最多 1 个字符）。
+angal.common.pleaseinsertacode.msg                                               = 请插入代码。
+
+angal.common.userid.label                                                        = 用户:
+angal.common.age.label                                                           = 年龄:
+angal.common.agefrom.label                                                       = Age From:
+angal.common.ageto.label                                                         = Age To:
+angal.common.datefrom.label                                                      = Date From:
+angal.common.dateto.label                                                        = Date To:
+angal.common.count.label                                                         = Count:
```
mwithi, line 85: Same doubts as in #585 (comment)

bundle/language_zh_CN.properties
```diff
@@ -1,1426 +1,1493 @@
...
+#  7. Module specific messages and text
+#  8. Text found in CORE repository
+#  9. Remaining original key/values that need review/work/etc.
+# 10. Key/values used in sql files (not changing them at this time to simplify the task)
+##########################################################
 
-angal.admission.abortdate                                                        = éåºæ¥æ
-angal.admission.address                                                          = å°å
-angal.admission.addressm                                                         = å°å
-angal.admission.admitted                                                         = æ¶æ²»
```
mwithi, line 5: was the original file not encoded properly (UTF-8)?

dbmalkovsky: Yes; that seems to be the issue.

PR #590, opened 2021-06-24
OP-542 Found a few English words left over

dbmalkovsky:
Strings with carriage returns are not always translated fully; this should fix that.


mwithi (approved):
Thanks a lot!

PR #607, opened 2021-06-26
OP-556 fix class not found exception with SkebbyGateway

dbmalkovsky:
See OP-556


dbmalkovsky:
@mwithi  I added `<scope>runtime</scope>`

PR #673, opened 2021-06-27
OP-557 add missing keys

dbmalkovsky:
See OP-557.


dbmalkovsky:
Instead of adding the keys back will change the values in the sql and load data files in Core project.   Closing this PR.

PR #787, opened 2021-07-02
OP-561 Improve StockCard dialog; add title and space

dbmalkovsky:
See OP-561

Attempted to make the least change yet make the dialog look better.

### DIFF
--- a/src/main/java/org/isf/utils/jobjects/JFromDateToDateChooser.java
+++ b/src/main/java/org/isf/utils/jobjects/JFromDateToDateChooser.java
@@ -21,6 +21,7 @@
  */
 package org.isf.utils.jobjects;
 
+import java.awt.FlowLayout;
 import java.awt.Font;
 import java.awt.LayoutManager;
 import java.beans.PropertyChangeEvent;
@@ -28,9 +29,7 @@ import java.beans.PropertyChangeListener;
 import java.util.Date;
 import java.util.Locale;
 
-import javax.swing.BoxLayout;
 import javax.swing.JLabel;
-import javax.swing.JOptionPane;
 import javax.swing.JPanel;
 
 import org.isf.generaldata.GeneralData;
@@ -41,7 +40,7 @@ import org.isf.generaldata.MessageBundle;
  */
 public class JFromDateToDateChooser extends JPanel {
 	
-	private static final String DATE_TIME_FORMAT = "dd/MM/yyyy";
+	private static final String DATE_TIME_FORMAT = " dd/MM/yyyy";
 	private static final int TEXT_SIZE = 12;
 	
 	private CustomJDateChooser dateFromDateChooser;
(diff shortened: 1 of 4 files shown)


dbmalkovsky:
The date from/to dialog is also used by the Stock Ledger option.

src/main/java/org/isf/utils/jobjects/JFromDateToDateChooser.java
```diff
@@ -41,7 +40,7 @@
  */
 public class JFromDateToDateChooser extends JPanel {
 	
-	private static final String DATE_TIME_FORMAT = "dd/MM/yyyy";
+	private static final String DATE_TIME_FORMAT = " dd/MM/yyyy";
```
mwithi, line 43: There was no other layout adjustment other than adding a space in the format?

mwithi: if I understand correctly the change...

mwithi (commented):
Just one question

src/main/java/org/isf/utils/jobjects/JFromDateToDateChooser.java
dbmalkovsky: Yes; I added a space so when it figures out the preferred spacing for the field there is actually enough room to see the whole date.  I was trying to make the most simple change in the code and this is what I came up with :)

mwithi: what about setting the preferred size?
`dateFromDateChooser.setPreferredSize(new Dimension(90, 20));`

dbmalkovsky: I was going to do that but at the time I considered that a "bigger" change that you might not approve.  Besides as shown in another PR figuring out the exact size can be tricky.

mwithi: > as shown in another PR figuring out the exact size can be tricky

It depends by the layout manager involved, some honor preferredSize() others ignore it and eventually require different approaches to obtain the wanted result. For the FlowLayout it is ok to use it, although it should guess the size by the JComponent, that in this case is a compound one, that's why I think it's tricky fo it to guess.

dbmalkovsky: So what is the final decision?  Leave the code as in this PR or replace it with `setPreferredSize()`?

mwithi: I think it is safe to use `setPreferredSize()` in this case (although there could be a better way to make the FlowLayout manager to calculate the rendering size by its own)

mwithi (approved):
Thanks a lot!

PR #791, opened 2021-07-05
OP-566 update dialog style to match

dbmalkovsky:
See OP-566

One minor change in the resource bundle for the title line and added ".toUpperCase()" for table column names.  No real code changes.


mwithi (approved):
Thanks!

PR #812, opened 2021-07-20
OP-583 Fix up Exams List window

dbmalkovsky:
See OP-583

Did a little formatting updating too.

### DIFF
--- a/src/main/java/org/isf/lab/gui/LabNew.java
+++ b/src/main/java/org/isf/lab/gui/LabNew.java
@@ -781,6 +781,7 @@ public class LabNew extends JDialog implements SelectionListener {
 					examPicker.setSize(300, 400);
 
 					JDialog dialog = new JDialog();
+					dialog.setTitle(MessageBundle.getMessage("angal.stat.examslist")); // TODO: use more correct key
 					dialog.setLocationRelativeTo(null);
 					dialog.setSize(600, 350);
 					dialog.setLocationRelativeTo(null);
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/lab/gui/LabNew.java
```diff
@@ -850,7 +851,7 @@ public void actionPerformed(ActionEvent e) {
 					
 					int selectedRow = jTableExams.getSelectedRow();
 					if (selectedRow < 0) {
-						MessageDialog.error(LabNew.this,"angal.labnew.pleaseselectanexam");
+						MessageDialog.error(LabNew.this,"angal.lab.pleaseselectanexam.msg");
```
mwithi, line 854: should we keep only the `angal.lab.` root?

src/main/java/org/isf/lab/gui/LabNew.java
```diff
@@ -781,6 +781,7 @@ public void actionPerformed(ActionEvent e) {
 					examPicker.setSize(300, 400);
 
 					JDialog dialog = new JDialog();
+					dialog.setTitle(MessageBundle.getMessage("angal.stat.examslist")); // TODO: use more correct key
```
mwithi, line 784: What in TODO that cannot be done now?

mwithi (commented):
LGTM. Just two comments to understand

src/main/java/org/isf/lab/gui/LabNew.java
dbmalkovsky: See note above; same reasoning.

src/main/java/org/isf/lab/gui/LabNew.java
dbmalkovsky: sure; but I was trying not to make resource bundle changes that might force Transifex to run.

mwithi: I see

src/main/java/org/isf/lab/gui/LabNew.java
mwithi: Ok

PR #817, opened 2021-07-21
OP-471 oh gui vaccine window error

mwithi:
See OP-471.

Also changed the whole layout and improved the model


src/main/java/org/isf/patvac/gui/PatVacEdit.java
```diff
@@ -58,6 +56,11 @@
 import org.isf.vaccine.model.Vaccine;
 import org.isf.vactype.manager.VaccineTypeBrowserManager;
 import org.isf.vactype.model.VaccineType;
+import java.awt.BorderLayout;
```
dbmalkovsky, line 59: These imports are out of order; they should be nearer to line 26

mwithi: Thanks!

src/main/java/org/isf/patvac/gui/PatVacEdit.java
```diff
@@ -513,27 +536,29 @@ private void setPatient(Patient selectedPatient) {
...
+			patientComboBox = new JComboBox<Object>();
+		}
+		patientComboBox.addItem(MessageBundle.getMessage("angal.patvac.selectapatient"));
 		Patient patSelected = null;
-		PatientBrowserManager patBrowser = Context.getApplicationContext().getBean(PatientBrowserManager.class);
 
 		if (GeneralData.ENHANCEDSEARCH){
 			try {
-				pat = patBrowser.getPatientsByOneOfFieldsLike(regExp);
+				patientList = patientManager.getPatientsByOneOfFieldsLike(regExp);
```
dbmalkovsky, line 548: should probably be spaces around the "catch"

mwithi: Done. Found only one...

PR #820, opened 2021-07-23
OP-343-release-package-naming

mwithi:
See OP-343.

Included .dist and .sample files to setup the application see README.md


README.md
```diff
@@ -11,24 +11,34 @@ To build this project you'll need Java JDK 8+ and Maven.
...
+(or use/modify `.sample` files with most common settings):
+
+`database.properties.dist 		-> database.properties` 	*(DBSERVER, DBPORT, DBNAME, DBUSER, DBPASS)*
+
+`dicom.properties.dist			-> dicom.properties`	 	*(OH_PATH_SUBSTITUTE/DICOM_DIR, DICOM_SIZE)*
+
+`log4j.properties.dist			-> log4j.properties` 		*(LOG_DEST, DBSERVER, DBPORT, DBNAME, DBUSER, DBPASS, LOG_LEVEL)*
+
+`settings.properties.dist		-> settings.properties`	*(OH_SET_LANGUAGE)*
+
```
pviotti, line 36:
```suggestion
| Dist file                | Property file       | Properties to fill in                                         |
|--------------------------|---------------------|---------------------------------------------------------------|
| database.properties.dist | database.properties | DBSERVER, DBPORT, DBNAME, DBUSER, DBPASS                      |
| dicom.properties.dist    | dicom.properties    | OH_PATH_SUBSTITUTE/DICOM_DIR, DICOM_SIZE                      |
| log4j.properties.dist    | log4j.properties    | LOG_DEST, DBSERVER, DBPORT, DBNAME, DBUSER, DBPASS, LOG_LEVEL |
| settings.properties.dist | settings.properties | OH_SET_LANGUAGE                                               |
```

pviotti: (courtesy of https://www.tablesgenerator.com/markdown_tables#)

pviotti (commented):
it feels a bit like an overkill to have `*.sample`, `*.dist` and the normal version of each file, but I might be missing some context here, and it it brings clarity for the users and the developers, all the better.

mwithi:
Thanks @pviotti, yes it's true, sample files are not needed if we assume that to set all those parameters in the table is not even more overkilling😁. Any suggestion to address that part so we can remove the .sample?

mwithi:
> and the normal version of each file

Removed database.properties that was not meant to be there, only not-generable .properties are kept

emecas (commented):
LGTM, but just wondering  what is the purpose of the "new" .dist files into the end to end workflow / installation process?

mwithi:
> LGTM, but just wondering what is the purpose of the "new" .dist files into the end to end workflow / installation process?

Thanks for the review. The .dist files are the "new" .ori files, they are meanth as templates to generate configuration files with the new startup scripts oh.sh/oh.ps1

mizzioisf (approved):

PR #823, opened 2021-07-23
OP-590 fix text truncation

dbmalkovsky:
See OP-590 for visual change.

Also cleaned the code up a little; had to do some reformatting unfortunately.


mwithi (approved):
Tested and LGTM. Thanks!

PR #827, opened 2021-07-26
Chore: instanceOf handles the null; if null then false

dbmalkovsky:
`instanceOf` returns false if the object is null so there is no need to check for it


mwithi (approved):
Thanks!

PR #828, opened 2021-07-26
OP-596: Update New Supplier window

dbmalkovsky:
See OP-596


mwithi (approved):
Tested and OK! Thanks!

PR #829, opened 2021-07-26
OP-594 Fix BillManager layout in singleuser mode

mwithi:
See OP-594.

Cc: @dbmalkovsky


dbmalkovsky:
Applied the PR and things LGTM in both single and multi-user modes.

PR #830, opened 2021-07-26
OP-597 Fix user selection upon new bill or new payment

mwithi:
See OP-597.


dbmalkovsky:
Applied the PR and things LGTM.

PR #834, opened 2021-07-28
OP-602 Settings->Exams->New/Edit dialog improvements

dbmalkovsky:
See OP-602


mwithi (approved):
Tested PR and it is OK. Thanks!

PR #836, opened 2021-07-28
OP-599 extend pex note field

mwithi:
See OP-599.

(build failing because it needs https://github.com/informatici/openhospital-core/pull/460)

Cc: @claudiorosazza

### DIFF
--- a/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
+++ b/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
@@ -887,7 +887,7 @@ public class PatientExaminationEdit extends ModalJFrame {
 
 	private VoLimitedTextArea getJTextAreaNote() {
 		if (jTextAreaNote == null) {
-			jTextAreaNote = new VoLimitedTextArea(PatientExamination.PEX_NOTE_LENGHT, 6, 20);
+			jTextAreaNote = new VoLimitedTextArea(PatientExamination.PEX_NOTE_LENGTH, 6, 20);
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -887,7 +887,7 @@ private CustomJDateChooser getJDateChooserDate() {
 
 	private VoLimitedTextArea getJTextAreaNote() {
 		if (jTextAreaNote == null) {
-			jTextAreaNote = new VoLimitedTextArea(300, 6, 20);
+			jTextAreaNote = new VoLimitedTextArea(2000, 6, 20);
```
dbmalkovsky, line 890: Would it make sense to have the value 2000 set as a public static variable in the persistence class and then used everywhere?    It would make the next change easier and ensure the exact same value is used.

mwithi: It is indeed also defined as JPA parameter:
```
	@Column(name="PEX_NOTE", length=2000)
	private String pex_note;
```
I was wondering if it can be used from there

mwithi: Done anyway, in combo with https://github.com/informatici/openhospital-core/pull/460

Thanks

dbmalkovsky:
If the corresponding code in CORE is fixed (This variable is spelled wrong; `LENGHT` should be `LENGTH`) this needs to be updated too.

PR #837, opened 2021-07-28
OP-604 Settings->Vaccine->New/Edit dialog improvements

dbmalkovsky:
Misc. code improvements and a little reformatting.

### DIFF
--- a/src/main/java/org/isf/vaccine/gui/VaccineEdit.java
+++ b/src/main/java/org/isf/vaccine/gui/VaccineEdit.java
@@ -76,25 +76,27 @@ public class VaccineEdit extends JDialog {
         vaccineListeners.remove(VaccineListener.class, listener);
     }
 
-    private void fireVaccineInserted() {
-        AWTEvent event = new AWTEvent(new Object(), AWTEvent.RESERVED_ID_MAX + 1) {
+	private void fireVaccineInserted() {
+		AWTEvent event = new AWTEvent(new Object(), AWTEvent.RESERVED_ID_MAX + 1) {
 
-			private static final long serialVersionUID = 1L;};
+			private static final long serialVersionUID = 1L;
+		};
 
-        EventListener[] listeners = vaccineListeners.getListeners(VaccineListener.class);
-	    for (EventListener listener : listeners)
-		    ((VaccineListener) listener).vaccineInserted(event);
-    }
+		EventListener[] listeners = vaccineListeners.getListeners(VaccineListener.class);
+		for (EventListener listener : listeners)
+			((VaccineListener) listener).vaccineInserted(event);
+	}
 
-    private void fireVaccineUpdated() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/vaccine/gui/VaccineEdit.java
```diff
@@ -131,10 +126,10 @@ public VaccineEdit(JFrame owner, Vaccine old, boolean inserting) {
...
-		pfrmHeight = 250;
-		pfrmBordX = screensize.width / 2 - (pfrmWidth / 2);
-		pfrmBordY = screensize.height / 2 - (pfrmHeight / 2);
+		int pfrmWidth = 350;
+		int pfrmHeight = 250;
+		int pfrmBordX = screensize.width / 2 - (pfrmWidth / 2);
+		int pfrmBordY = screensize.height / 2 - (pfrmHeight / 2);
 		this.setBounds(pfrmBordX,pfrmBordY,pfrmWidth,pfrmHeight);
 		this.setContentPane(getJContentPane());
 		if (insert) {
```
mwithi, line 135: In many cases we just replaced all this stuff with:
```
		pack();
		setLocationRelativeTo(null);
```

`pack()` allows to double check the minimum size and to check the good implementation of the layout.
`setLocationRelativeTo(null)` allows to center the result anyhow

Can you give it a try?

src/main/java/org/isf/vaccine/gui/VaccineEdit.java
```diff
@@ -166,28 +161,26 @@ private JPanel getJContentPane() {
...
+			JLabel codeLabel = new JLabel(MessageBundle.getMessage("angal.common.code.txt") + ':');
 			// vaccine description
-			descLabel = new JLabel(MessageBundle.getMessage("angal.common.description.txt"));
+			JLabel descLabel = new JLabel(MessageBundle.getMessage("angal.common.description.txt") + ':');
 			// required fields
-			requiredLabel= new JLabel();
-			requiredLabel.setText(MessageBundle.getMessage("angal.vaccine.requiredfields"));
-			
+			JLabel requiredLabel = new JLabel(MessageBundle.getMessage("angal.vaccine.requiredfields"));
+
```
mwithi, line 172: It seems requiredLabel is not needed here, or can be changed to "all fields required", but if so also similar windows have same problem.

mwithi (changes requested):
"Required field" is not needed and layout can be automated

src/main/java/org/isf/vaccine/gui/VaccineEdit.java
mwithi: Thanks!

src/main/java/org/isf/vaccine/gui/VaccineEdit.java
mwithi: Thanks!

mwithi (approved):

PR #851, opened 2021-08-06
OP-621 Settings->Types->Other Prices and New/Edit

dbmalkovsky:
See OP-621

### DIFF
--- a/src/main/java/org/isf/pricesothers/gui/PricesOthersEdit.java
+++ b/src/main/java/org/isf/pricesothers/gui/PricesOthersEdit.java
@@ -23,9 +23,6 @@ package org.isf.pricesothers.gui;
 
 import java.awt.AWTEvent;
 import java.awt.BorderLayout;
-import java.awt.GridLayout;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.util.EventListener;
 
 import javax.swing.BoxLayout;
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/pricesothers/gui/PricesOthersEdit.java
```diff
@@ -46,51 +44,51 @@
 import org.isf.utils.exception.gui.OHServiceExceptionUtil;
 import org.isf.utils.jobjects.MessageDialog;
 import org.isf.utils.jobjects.VoLimitedTextField;
+import org.isf.utils.layout.SpringUtilities;
 
 public class PricesOthersEdit extends JDialog {
 	
-	private EventListenerList OthersListeners = new EventListenerList();
+	private EventListenerList othersListeners = new EventListenerList();
```
mwithi, line 51: Might we refactor into `PricesOthersListeners` ?

mwithi (approved):
LGTM. Just one comment

src/main/java/org/isf/pricesothers/gui/PricesOthersEdit.java
dbmalkovsky: Yes we can.  Note it should start with a lowercase character.  I went ahead and changed method names to be more clear.

mwithi: sure, thanks!

PR #853, opened 2021-08-14
Chore: fix spelling

dbmalkovsky:
It appears that this property isn't really being used but if it is, correct the spelling.


mwithi (approved):
Thanks!

PR #855, opened 2021-08-16
OP-453 Remove unused keys

dbmalkovsky:
See OP-453

These unused keys were found after running a script to match keys with usage.

There are only two remaining keys that appear to be used in `step_05_update_menu_i18n.sql` that do not exist in a properties file:

- `angal.menu.laboratoryresulttype`
- `angal.menu.btn.laboratoryresulttype`

My searching indicates that they are no longer part of the product.   If that is true then after this commit the issue can be closed.


mwithi (approved):
LGTM. Thanks a lot!

PR #859, opened 2021-08-18
OP-626_close-button-in-patient-examination

mwithi:
See OP-626.

### DIFF
--- a/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
+++ b/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
@@ -1190,14 +1190,13 @@ public class PatientExaminationEdit extends ModalJFrame {
 			jSliderHeight = new JSlider(ExaminationParameters.HEIGHT_MIN, ExaminationParameters.HEIGHT_MAX, ExaminationParameters.HEIGHT_INIT);
 			jSliderHeight.addChangeListener(e -> {
 				JSlider source = (JSlider)e.getSource();
-		        if (!source.getValueIsAdjusting()) {
-		        	System.out.println("change!");
-		        	int value = jSliderHeight.getValue();
+				if (!source.getValueIsAdjusting()) {
+					int value = jSliderHeight.getValue();
 					jTextFieldHeight.setText(String.valueOf(value));
 					patex.setPex_height(value);
 					updateBMI();
 					modified = true;
-		        }    
+				}    
 			});
 			jSliderHeight.setFocusable(false);
 		}
@@ -1208,15 +1207,14 @@ public class PatientExaminationEdit extends ModalJFrame {
 		if (jSliderWeight == null) {
 			jSliderWeight = new ScaledJSlider(ExaminationParameters.WEIGHT_MIN, ExaminationParameters.WEIGHT_MAX, ExaminationParameters.WEIGHT_STEP, ExaminationParameters.WEIGHT_INIT);
 			jSliderWeight.addChangeListener(e -> {
-				JSlider source = (JSlider)e.getSource();
-		        if (!source.getValueIsAdjusting()) {
-		        	System.out.println("change!");
-		        	double value = jSliderWeight.getScaledValue();
+				JSlider source = (JSlider) e.getSource();
+				if (!source.getValueIsAdjusting()) {
+					double value = jSliderWeight.getScaledValue();
 					jTextFieldWeight.setText(String.valueOf(value));
 					patex.setPex_weight(value);
 					updateBMI();
 					modified = true;
-		        }
+				}
 			});
 			jSliderWeight.setFocusable(false);
 		}
--- a/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
+++ b/src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
@@ -139,7 +139,7 @@ public class PatientExaminationEdit extends ModalJFrame {
 	private JComboBox jComboBoxAuscultation;
 	private JButton jButtonSave;
 	private JButton jButtonDelete;
-	private JButton jButtonCancel;
+	private JButton jButtonClose;
 	private JButton jButtonPrint;
 	private Action actionSavePatientExamination;
 	private Action actionToggleAP;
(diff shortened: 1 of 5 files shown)


src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1190,10 +1189,15 @@ private JSlider getJSliderHeight() {
 		if (jSliderHeight == null) {
 			jSliderHeight = new JSlider(ExaminationParameters.HEIGHT_MIN, ExaminationParameters.HEIGHT_MAX, ExaminationParameters.HEIGHT_INIT);
 			jSliderHeight.addChangeListener(e -> {
-				int value = jSliderHeight.getValue();
-				jTextFieldHeight.setText(String.valueOf(value));
-				patex.setPex_height(value);
-				updateBMI();
+				JSlider source = (JSlider)e.getSource();
+		        if (!source.getValueIsAdjusting()) {
+		        	System.out.println("change!");
```
dbmalkovsky, line 1194: Not sure you want to leave this in production.

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1204,11 +1208,15 @@ private ScaledJSlider getJSliderWeight() {
...
 			jSliderWeight = new ScaledJSlider(ExaminationParameters.WEIGHT_MIN, ExaminationParameters.WEIGHT_MAX, ExaminationParameters.WEIGHT_STEP, ExaminationParameters.WEIGHT_INIT);
 			jSliderWeight.addChangeListener(e -> {
-				double value = jSliderWeight.getScaledValue();
-				//double weight = (double) value / 10;
-				jTextFieldWeight.setText(String.valueOf(value));
-				patex.setPex_weight(value);
-				updateBMI();
+				JSlider source = (JSlider)e.getSource();
+		        if (!source.getValueIsAdjusting()) {
+		        	System.out.println("change!");
```
dbmalkovsky, line 1213: And here too

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1190,10 +1189,15 @@ private JSlider getJSliderHeight() {
...
 			jSliderHeight.addChangeListener(e -> {
-				int value = jSliderHeight.getValue();
-				jTextFieldHeight.setText(String.valueOf(value));
-				patex.setPex_height(value);
-				updateBMI();
+				JSlider source = (JSlider)e.getSource();
+		        if (!source.getValueIsAdjusting()) {
+		        	System.out.println("change!");
+		        	int value = jSliderHeight.getValue();
+					jTextFieldHeight.setText(String.valueOf(value));
```
dbmalkovsky, line 1196: formatting?

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1204,11 +1208,15 @@ private ScaledJSlider getJSliderWeight() {
...
-				double value = jSliderWeight.getScaledValue();
-				//double weight = (double) value / 10;
-				jTextFieldWeight.setText(String.valueOf(value));
-				patex.setPex_weight(value);
-				updateBMI();
+				JSlider source = (JSlider)e.getSource();
+		        if (!source.getValueIsAdjusting()) {
+		        	System.out.println("change!");
+		        	double value = jSliderWeight.getScaledValue();
+					jTextFieldWeight.setText(String.valueOf(value));
```
dbmalkovsky, line 1215: formatting?

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
mwithi: Sure not! thanks!

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
mwithi: Thanks!

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
mwithi: Thanks!

dbmalkovsky:
I believe the "modification indicator" needs additional logic (and this is why I had suggested we don't do it at this time).   For example, if I click one of the check boxes, say next to "Heart Rate", and then click the same checkbox again to uncheck it without making any other modifications, when I click on "Close" it asks me whether I want to save the changes.   There aren't really any differences.

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -65,6 +65,7 @@
 import javax.swing.OverlayLayout;
 import javax.swing.ScrollPaneConstants;
 import javax.swing.SwingConstants;
+import javax.swing.Timer;
```
dbmalkovsky, line 68: I don't believe this import is used.

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1330,11 +1336,24 @@ private JButton getJButtonDelete() {
 		return jButtonDelete;
 	}
 	
-	private JButton getJButtonCancel() {
+	private JButton getJButtonClose() {
 		if (jButtonCancel == null) {
```
dbmalkovsky, line 1340: shouldn't the button be renamed to `jButtonClose`?

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
mwithi: Thanks!

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
mwithi: Yep, thank you!

mwithi:
> I believe the "modification indicator" needs additional logic (and this is why I had suggested we don't do it at this time). For example, if I click one of the check boxes, say next to "Heart Rate", and then click the same checkbox again to uncheck it without making any other modifications, when I click on "Close" it asks me whether I want to save the changes. There aren't really any differences.

I know about this, I've tried to implement more rigorous logics but the way fields are updated doesn't make things simple. Anyway I think the current logic it's a good compromise for a reasonable usage: how many users will just check and uncheck one parameter only (as the only change) and expecting not to be warned about changes? In contrary, to have at least a check is what a common user would expect. We have to consider real cases... majority will just open the window, set parameters and save... do you agree?

mwithi:
> Yes that is the typical usage.

approved?

dbmalkovsky:
LGTM given the limitation.

PR #870, opened 2021-08-22
OP-630: fixing rendering issue on Therapy dialog

alexwibowo:
Need to call pack() on the dialog before showing it, especially when we are adding components to the panel / dialog.
This will cause the window to be calculated & painted properly. Otherwise some components wont be shown properly.


mwithi (approved):
LGTM. Thanks!

PR #872, opened 2021-08-23
OP-153 lot information at the wards level fixes

mwithi:
Small fixes related to OP-153.


src/main/java/org/isf/medicalstock/gui/MovStockMultipleDischarging.java
```diff
@@ -958,13 +946,24 @@ private boolean checkAndPrepareMovements() {
 			mov.setWard((Ward) jComboBoxDestination.getSelectedItem());
 			mov.setDate(thisDate);
 			mov.setRefNo(jTextFieldReference.getText());
+			quantities.add(mov.getQuantity());
```
dbmalkovsky, line 949: why is `quantities` an `ArrayList<Integer>` when `getQuantity()` returns an `int`?  Make the array the same type so there is no boxing required.

src/main/java/org/isf/medicalstock/gui/MovStockMultipleDischarging.java
```diff
@@ -958,13 +946,24 @@ private boolean checkAndPrepareMovements() {
...
 		return ok;
 	}
 	
+	private void rollBackMovements() {
+		ArrayList<Movement> movements = model.getMovements();
+
+		// Set back changed quantities
+		for (int i = 0; i < movements.size(); i++) {
+			Movement mov = movements.get(i);
+			mov.setQuantity(quantities.get(i));
```
dbmalkovsky, line 963: here is another place that boxing would be required if not changed to `int`.

dbmalkovsky:
Other than the comment; LGTM.

src/main/java/org/isf/medicalstock/gui/MovStockMultipleDischarging.java
mwithi: @dbmalkovsky could you please suggest the code modification directly?

dbmalkovsky: Nevermind; I was wrong.

PR #874, opened 2021-08-23
Chore: "static" base class members should not be accessed via derived types

dbmalkovsky:
Use the base class that actually contains the constant and not various subclasses; can cause confusion.


mwithi (approved):
LGTM, thanks!

PR #879, opened 2021-09-13
Chore: remove duplicate code in both branches of if

dbmalkovsky:
Move duplicate code in both branches of `if` out of the conditional.

A little formatting ended up in the PR too.


mwithi (approved):
LGMT. Thanks!

PR #881, opened 2021-09-13
Chore: use bulk method instead of iteration for performance

dbmalkovsky:
What started out as just using a bulk method, `addAll()` turned into removing a unused variable and fixing the name of another class variable so it didn't start with an uppercase `J`.

### DIFF
--- a/src/main/java/org/isf/medstockmovtype/gui/MedicalsrMovPatList.java
+++ b/src/main/java/org/isf/medstockmovtype/gui/MedicalsrMovPatList.java
@@ -50,8 +50,10 @@ public class MedicalsrMovPatList extends JPanel {
 	private JTable jTableData;
 	private OhTableDrugsModel<MovementWard> modelMedWard;
 	private OhDefaultCellRenderer cellRenderer = new OhDefaultCellRenderer();
+	private MovWardBrowserManager movManager = Context.getApplicationContext().getBean(MovWardBrowserManager.class);
+
 	public MedicalsrMovPatList(Object object) {
-		
+
 		setLayout(new BorderLayout(0, 0));
 		JPanel panelData = new JPanel();
 		add(panelData);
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medstockmovtype/gui/MedicalsrMovPatList.java
```diff
@@ -47,10 +47,9 @@
 	private Patient myPatient;
 	private ArrayList<MovementWard> drugsData;
 	private JDialog dialogDrug;
-	private JTable JtableData;
+	private JTable jTableData;
 	private OhTableDrugsModel<MovementWard> modelMedWard;
 	private OhDefaultCellRenderer cellRenderer = new OhDefaultCellRenderer();
-	private MovWardBrowserManager movManager = Context.getApplicationContext().getBean(MovWardBrowserManager.class);
```
mwithi, line 53: I find it is better to have all managers loaded at the begin of the class, wdyt?

dbmalkovsky: Sure if it is being used but it appears not in this case so it is just a performance hit, isn't it?

mwithi: it is, at line 71, no?

mwithi: I mean, it was enough to remove line 68

dbmalkovsky: Understand now; fixed.

mwithi: alright! thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #1059, opened 2021-12-07
OP-720 fix date sorting

dbmalkovsky:
See OP-720

Use simple date format of "dd/MM/yy" for sorting the date column instead of the `toString()` of `date`.

Also moved the `sortByColumn()` instruction to line up better with similar code in `PatientFolderBrowser`.


mwithi (commented):
As mentioned in OP-720, the "solution" could be to fix TableSorter rather than classes that use it, because the expected result is that the sorter should sort columns taking into account any type (and not only String).

Otherwise LGTM, same fix is coming soon from [hotfix](https://github.com/informatici/openhospital-gui/blob/hotfix/src/main/java/org/isf/admission/gui/PatientDataBrowser.java) branch.

mwithi:
Created a more specific issue for this in OP-723.

PR #1078, opened 2021-12-24
Chore: update pom.xml based on dependabot PR

dbmalkovsky:
I will rebase the dependabot PR after this is comitted.


mwithi (approved):
Thanks!

PR #1082, opened 2021-12-29
Reworked startup script configuration

mizzioisf:
Final updates for 1.11.2
- Reworked configuration generation: configuration files are generated at the first startup and not overwritten, unless the option GENERATE_CONFIG_FILES is set to on
- Better checks
- Clearer console and log output 
- Added the possibility of setting single or multi user OH environment (SINGLEUSER)
Needs testing!


oh.bat
```diff
@@ -122,11 +122,12 @@ set DATABASE_USER=isf
 set DATABASE_PASSWORD=isf123
 
 set DICOM_MAX_SIZE="4M"
-set DICOM_STORAGE="FileSystemDicomManager" REM SqlDicomManager
+set DICOM_STORAGE="FileSystemDicomManager"
 set DICOM_DIR="data/dicom_storage"
 
-set OH_DIR="."
+set OH_DIR="oh"
```
mwithi, line 128: this should be set back to "." like the others?

mizzioisf: Yes right, corrected.

mwithi (approved):
LGTM!

PR #1084, opened 2021-12-30
Merge Hotfix v1.11.1 after release

mwithi:


src/main/java/org/isf/visits/gui/InsertVisit.java
```diff
@@ -223,6 +225,37 @@ private JPanel getWardPanel() {
 		return wardPanel;
 	}
 
+	private JComboBox<Ward> getWardBox() {
```
dbmalkovsky, line 228: Look starting at line 259, the method getWardBox() is already defined in this class.    One of them needs to be deleted to  compile properly.

mwithi: I was sure to have deleted one of the two... thanks!

PR #1087, opened 2021-12-31
OP-737 make month values match better

dbmalkovsky:
See OP-737


mwithi (changes requested):
Good fix! Anyway it is not the only place where `JMonthChooser` is used in the gui, can you apply also to other cases?

dbmalkovsky:
The latest commit also fixed an issue with trying to edit a therapy when setting the number of days, weeks, and months.   The fix eliminated another JODA time usage.

dbmalkovsky:
@mwithi I think I addressed all the changes (and then some).

mwithi (approved):
Tested and LGTM.

Anyway while testing another issue raised up OP-745.

PR #1088, opened 2022-01-01
OP-332 consolidate JasperReports dependency in Core

dbmalkovsky:
See OP-332 and [Core PR-560](https://github.com/informatici/openhospital-core/pull/560
)


mwithi (approved):
Thanks!

PR #1102, opened 2022-01-20
OP-751 fix date pattern

dbmalkovsky:
See OP-751

Companion to CORE PR https://github.com/informatici/openhospital-core/pull/586


mwithi (approved):
LGTM. Thanks!

PR #1109, opened 2022-01-29
OP-748 Use abstract method instead of concrete method

dbmalkovsky:
See OP-748


mwithi (commented):
The change fixes part of the issue, that states:

"Other values should not be managed (it also always shows OPD as fallback value), nor accepted by the manager, nor visualized."

dbmalkovsky:
Can someone please explain in more detail what this means?
<pre>
"Other values should not be managed (it also always shows OPD as fallback value), nor accepted by the manager, nor visualized."
</pre>

mwithi:
> Can someone please explain in more detail what this means?

I've updated the issue description OP-748.

PR #1143, opened 2022-02-28
Chore: use date constant strings in org.isf.utils.Constants instead of strings or local constants

dbmalkovsky:
Think this should be the last of the hard coded date/time string patterns.


mwithi (approved):
LGTM. Thanks!

PR #1146, opened 2022-03-08
OP-821 Fix calculation of last day of the month.

dbmalkovsky:
See OP-821

### DIFF
--- a/src/main/java/org/isf/medicals/gui/MedicalBrowser.java
+++ b/src/main/java/org/isf/medicals/gui/MedicalBrowser.java
@@ -646,9 +646,9 @@ public class MedicalBrowser extends ModalJFrame implements MedicalListener {
 	}
 
 	private LocalDate getToDatePlusMonth(int monthsToMove) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicals/gui/MedicalBrowser.java
```diff
@@ -646,9 +646,9 @@ protected void launchExpiringReport() {
 	}
 
 	private LocalDate getToDatePlusMonth(int monthsToMove) {
-		return LocalDate.now()
-				.plusMonths(monthsToMove)
-				.withDayOfMonth(LocalDate.now().lengthOfMonth());
+		LocalDate plusMonth = LocalDate.now().plusMonths(monthsToMove);
+		return plusMonth
+				.withDayOfMonth(plusMonth.lengthOfMonth());
```
mwithi, line 651:
```suggestion
		return plusMonth.withDayOfMonth(plusMonth.lengthOfMonth());
```

PR #1148, opened 2022-03-10
Chore update to slf4j-reload4j

dbmalkovsky:
Change slf4j-log4j12 to slf4j-reload4j which is the new name; see Reload4j 1.2.18 as a replacement for log4j 1.2.17 at https://www.slf4j.org/log4shell.html


mwithi (approved):
Thanks!!!

PR #1155, opened 2022-03-21
OP-826 fix parameter name

dbmalkovsky:
See OP-826  misspelled paramete


dbmalkovsky:
@mwithi I'm assuming that the jasper file needs to be regenerated too.  Is that something you can do?  Thanks

mwithi:
> @mwithi I'm assuming that the jasper file needs to be regenerated too. Is that something you can do? Thanks

yep! I will recompile (with Jasper Studio)

PR #1171, opened 2022-05-03
OP-860 update OPD dialogs

dbmalkovsky:
See OP-860


mwithi (approved):
LGTM. Thanks!

src/main/java/org/isf/opd/gui/OpdBrowser.java
```diff
@@ -379,7 +379,9 @@ private JButton getJDeleteButton() {
 				if (GeneralData.OPDEXTENDED) {
 					message = MessageBundle.formatMessage("angal.opd.deletefollowingopdextended.fmt.msg",
 							opd.getPatient().getName(),
-							opd.getCreatedDate().format(DATE_TIME_FORMATTER),
+							opd.getCreatedDate() == null
+									? opd.getDate().format(DATE_FORMATTER)
+									: opd.getCreatedDate().format(DATE_FORMATTER),
```
mwithi, line 384: Why to take into account the creation date OR the visit date? The latter should be always notnull and the former only for audit purpose

PR #1173, opened 2022-05-09
Chore: add spinner option to time picker; do not change specified time

dbmalkovsky:
Add spinner option to time picker.
Fix issue where the minute, second, and nanosecond values were zeroed out.


mwithi (approved):
LGTM. Thanks

PR #1174, opened 2022-05-10
OP-866 remove time from Lot Information dialog

dbmalkovsky:
Time was added by mistake in previous push.


mwithi (approved):
Thanks a lot @dbmalkovsky !

PR #1178, opened 2022-05-17
Chore: remove GregorianCalendar usage

dbmalkovsky:
This is when creating a new OPD entry when `OPDEXTENDED=no` in the settings file

### DIFF
--- a/src/main/java/org/isf/opd/gui/OpdEdit.java
+++ b/src/main/java/org/isf/opd/gui/OpdEdit.java
@@ -31,7 +31,6 @@ import java.awt.event.FocusEvent;
 import java.awt.event.FocusListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.KeyListener;
-import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.EventListener;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/opd/gui/OpdEdit.java
```diff
@@ -508,8 +507,7 @@ private JButton getOkButton() {
 
 						try {
 							if (insert) {    // Insert
-								GregorianCalendar date = new GregorianCalendar();
-								opd.setProgYear(opdManager.getProgYear(date.get(Calendar.YEAR)) + 1);
+								opd.setProgYear(opdManager.getProgYear(LocalDate.now().getYear()) + 1);
```
mwithi, line 510: In the case OPDEXTENDED=yes we call .getProgYear() method passing the visitDate (so it calculates it within the specified year)

dbmalkovsky: Well that is a change from the behavior that I modified here isn't it?   It was using "NOW" and I just did the same.
This is a NEW  (ie, insert) so is there even a visit date?

mwithi: if you like you can blame the forgotten change among OP-719 (my bad)

dbmalkovsky: No need to find blame; I've updated the code; please review @mwithi

mwithi: Thanks!

PR #1181, opened 2022-05-18
OP-874 show time for Admission and Discharge

dbmalkovsky:
See OP-874


mwithi (approved):
LGTM and tested. Thanks!

PR #1190, opened 2022-06-01
OP-880 Delivery panel changes for date/time pickers

dbmalkovsky:
See OP-880

### DIFF
--- a/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
+++ b/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
@@ -221,13 +221,13 @@ public class AdmissionBrowser extends ModalJFrame {
 
 	private JComboBox treatmTypeBox = null;
 
-	private static final int preferredWidthDates = 225;
+	private static final int PREFERRED_WIDTH_DATES = 225;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/admission/gui/AdmissionBrowser.java
```diff
@@ -221,21 +217,21 @@ private void fireAdmissionUpdated(Admission anAdmission) {
...
+	private static final int preferredWidthDates = 225;
 
-	private final int preferredWidthDiagnosis = 550;
+	private static final int preferredWidthDiagnosis = 550;
 
-	private final int preferredWidthTypes = 220;
+	private static final int preferredWidthTypes = 220;
 
-	private final int preferredHeightLine = 24;
+	private static final int preferredHeightLine = 24;
```
mwithi, line 230: should we capitalize these?

dbmalkovsky: Done

mwithi: Thanks a lot!

PR #1196, opened 2022-06-07
OP-820 add hospital level visit start/end time to browser

dbmalkovsky:
See OP-820

Settings -> Hospital

This depends on the Core PR: https://github.com/informatici/openhospital-core/pull/678

I simplified the layout code for the Hospital Browser window in the process.


mwithi (approved):
LGTM. But is this PR still a WIP?

dbmalkovsky:
I think we can merge this so as to move forward with ward level changes.

PR #1205, opened 2022-06-15
OP-820 add hospital level visit duration to Hospital browser

dbmalkovsky:
continue work on OP-820


mwithi (approved):
LGMT. Tested also. Thanks!

PR #1209, opened 2022-06-18
OP-913 Change date/time widget in Lab->New/Edit when LABEXTENDED is yes and MULTIPLEINSERT is no

dbmalkovsky:
See OP-913

Also made some small code improvements found by SonarLint and Intellij Analyze


mwithi (changes requested):
The "Date" field in "New" window shoul be pre-filled
(with actual date and time or from `RememberDates`)

image

mwithi (approved):

PR #1213, opened 2022-06-23
OP-704 Add search box for Operations in OPD and Admission tabs

nlazzer1:
The current operation drop down list has been integrated with a filter
component (textbox + search button)

See OP-704.


mwithi (approved):
LGTM. Thanks!

PR #1224, opened 2022-07-04
OP-920 standardize year formatting and use LocalDateTime objects directly

dbmalkovsky:
See OP-920


dbmalkovsky:
Sorry about the reformatting; there were multiple styles there and I couldn't get them to agree without reformatting the whole block.

mwithi (approved):
LGTM. Thanks!

PR #1234, opened 2022-07-17
Op 312 - telemetry module

AndreiDodu:
- merged develop
- changed FreeGeoIPJSON -> GeoIpLookup service


src/main/java/org/isf/telemetry/gui/TelemetryGUI.java
```diff
@@ -37,7 +37,7 @@
```
dbmalkovsky, line 37: Isn't adding `GUI` to the end of the package name redundant as it is in the GUI project, the package name also includes `gui` and it is built into the gut jar?     That may have made sense years ago when everything was in a single tree/repo.  I would really like to get rid of all such instances in this project.

PR #1242, opened 2022-07-22
OP-944 Change date choosers for Lot Information

dbmalkovsky:
See OP-944

Did some other file cleanup based on SonarLint and IntelliJ Analyze tools

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
@@ -76,6 +76,7 @@ import org.isf.utils.exception.gui.OHServiceExceptionUtil;
 import org.isf.utils.jobjects.GoodDateChooser;
 import org.isf.utils.jobjects.MessageDialog;
 import org.isf.utils.jobjects.RequestFocusListener;
+import org.isf.utils.jobjects.TextPrompt;
 import org.isf.utils.time.TimeTools;
 import org.isf.ward.model.Ward;
 
@@ -536,6 +537,14 @@ public class WardPharmacyRectify extends JDialog {
 		LocalDateTime expiringDate = LocalDateTime.now();
 		Lot lot = null;
 		JTextField lotNameTextField = new JTextField(15);
+
+		TextPrompt suggestion = new TextPrompt(MessageBundle.getMessage("angal.medicalstock.multiplecharging.lotid"), lotNameTextField);
+		suggestion.setFont(new Font("Tahoma", Font.PLAIN, 14)); //$NON-NLS-1$
+		suggestion.setForeground(Color.GRAY);
+		suggestion.setHorizontalAlignment(SwingConstants.CENTER);
+		suggestion.changeAlpha(0.5f);
+		suggestion.changeStyle(Font.BOLD + Font.ITALIC);
+
 		lotNameTextField.addAncestorListener(new RequestFocusListener());
 		if (isAutomaticLot()) {
 			lotNameTextField.setEnabled(false);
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -727,12 +699,10 @@ private JComboBox getJComboBoxMedical() {
 					Medical med = ((Medical) jComboBoxMedical.getSelectedItem());
 					jButtonChooseLot.setEnabled(false);
 					jButtonNewLot.setEnabled(true);
-					lotExist = false;
 					for (MedicalWard medWard : wardDrugs) {
 						if (med.getDescription().equals(medWard.getMedical().getDescription())) {
 							jButtonChooseLot.setEnabled(true);
 							jButtonNewLot.setEnabled(false);
-							lotExist = true;
```
mwithi, line 735: I've doubts concerning this removal... maybe we lost some logic back in date?

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -198,266 +195,243 @@ public WardPharmacyRectify(JFrame owner, Ward ward, Medical medical) {
...
+
+		JButton jButtonOk = new JButton(MessageBundle.getMessage("angal.common.ok.btn"));
+		jButtonOk.setMnemonic(MessageBundle.getMnemonic("angal.common.ok.btn.key"));
+		jButtonOk.addActionListener(actionEvent -> {
+			Object item;
+			Medical med;
+
 			/*
-			 * TODO: to refactor all this part by extracting in separated method all this logic
+			 *  To override MovWardBrowserManager.validateMovementWard() behavior
```
mwithi, line 333: Why to remove the TODO reminder?

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
dbmalkovsky: There are only three occurrences of `lotExist`; one is the definition and the other two are here.  The value is never tested so the assignment are not needed.   You added the code on June 12, 2020 and as far as I can tell looking at other versions between now and then nothing was ever done with the variable.

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
dbmalkovsky: because it is redundant; see line 329

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
mwithi: I can see that, but I'm wandering if there was any logic before that time... maybe we need to verify that part: the JComboBox should show only medicals with a lot whose quantity is different from zero. 
Maybe this check is done elsewhere in the same class?

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
mwithi: ok, but you also changed the comment, so they are 2 different TODOs to take care?

dbmalkovsky: I put the comment back

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
dbmalkovsky: I checked all 26 commits between now and when the code as added on June 12, 2020.  There are always 3 occurrences of `lotExist`.  The private definition of the variable; the set to false; and the set to true.

I then checked all 12 commits before June 12, 2020 and none of them had the variable.

It looks like it was an incomplete thought that is lost in time.....

mwithi: yes, could be! I will test it deeply... thanks!

dbmalkovsky:
Add watermark "Lot id" as requested.

mwithi (approved):
LGTM. Thanks!

PR #1243, opened 2022-07-22
addendum OP-312 telemetry module

AndreiDodu:


src/main/java/org/isf/telemetry/gui/TelemetryGUI.java
```diff
@@ -84,17 +85,13 @@ public class TelemetryGUI extends JDialog {
 	private JPanel panel;
 	private TelemetryManager telemetryManager = Context.getApplicationContext().getBean(TelemetryManager.class);
 	private TelemetryUtils telemetryUtils = Context.getApplicationContext().getBean(TelemetryUtils.class);
-
-	public TelemetryGUI() {
-		super();
-		init();
-	}
```
mwithi, line 91: I think this constructor is needed to open the window from Main Menu > Settings, could you please verify?

rsc/telemetry.properties
```diff
@@ -3,4 +3,7 @@ telemetry.daemon.enabled=true
 # number of second to check if it should send a message
 telemetry.daemon.thread.loop.seconds=10
 # remote data collector server base url
-telemetry-gateway-service.base-url=http://localhost:8080
\ No newline at end of file
+telemetry-gateway-service.base-url=http://localhost:8080
+
+
+geoiplookup-remote-service.ribbon.base-url=https://json.geoiplookup.io
```
mwithi, line 9: I would add a comment here.

Is there any chance to generalize the parameter?

mwithi (changes requested):
I think the emtpy constructor is needed anyway
Also comment in properties is requested
Thanks!

mwithi:
Superseded by https://github.com/informatici/openhospital-gui/pull/1265

PR #1250, opened 2022-07-28
OP-941 do not allow changed OPD No.

dbmalkovsky:
See OP-941

Sorry about the automatic reformat on the save of the file.


mwithi (approved):
Very good! Thanks a lot!

PR #1251, opened 2022-07-28
OP-942 fix locking exception

dbmalkovsky:
See OP-942

Note: having a hard time testing this in my current environment.


mwithi (approved):
LGTM. Thanks!

PR #1254, opened 2022-07-30
OP-939 show all wards; move validation to core; set dialog title

dbmalkovsky:
See OP-939

Requires PR in CORE: https://github.com/informatici/openhospital-core/pull/720


mwithi (approved):
LGTM. Thanks!

PR #1255, opened 2022-07-31
OP-704-FIX:

nlazzer1:
Fix bug when searching for an intervention (changed the size of
components based on the length of the interventions) (See OP-704)

### DIFF
--- a/src/main/java/org/isf/operation/gui/OperationRowBase.java
+++ b/src/main/java/org/isf/operation/gui/OperationRowBase.java
@@ -102,8 +102,8 @@ abstract class OperationRowBase extends JPanel {
 
 		add(panelForm, BorderLayout.NORTH);
 		GridBagLayout gbl_panelForm = new GridBagLayout();
-		gbl_panelForm.columnWidths = new int[] { 60, 75, 0, 300, 0, 0 };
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/operation/gui/OperationRowBase.java
```diff
@@ -117,12 +117,14 @@ protected OperationRowBase() {
...
 		searchOperationTextField = new JTextField();
+		searchOperationTextField.setPreferredSize(new Dimension(150, 20));
 		GridBagConstraints gbc_searchOperation = new GridBagConstraints();
+		gbc_searchOperation.anchor = GridBagConstraints.WEST;
 		gbc_searchOperation.insets = new Insets(0, 0, 5, 5);
 		gbc_searchOperation.fill = GridBagConstraints.HORIZONTAL;
 		gbc_labelOperation.gridx = 1;
 		gbc_labelOperation.gridy = 0;
-		searchOperationTextField.setColumns(20);
+		//searchOperationTextField.setColumns(120);
```
mwithi, line 127: no need to keep previous code commented (we use github logs in case)

mwithi (approved):
Tested and LGTM. Only remove one commented line of code if possible

PR #1258, opened 2022-07-31

dbmalkovsky:
See OP-943

This requires CORE PR https://github.com/informatici/openhospital-core/pull/723 to be merged to compile cleanly


mwithi (approved):
LGTM. Thanks!

PR #1259, opened 2022-07-31

dbmalkovsky:
This is the hotfix version

See OP-943


mwithi (approved):
LGMT. Thanks!

PR #1261, opened 2022-08-01

dbmalkovsky:
See OP-943

Fixes the `contains()` test by using the longer string to test; `.xls` is also a part of `.xlsx`.   Sorry.


mwithi (approved):
😄 agreed! Thanks!

PR #1264, opened 2022-08-02
OP-949 fix classcastexception during Recity

dbmalkovsky:
See OP-949

I don't like the fix but it works.    Depending on whether or not the drug has any stock and/or whether the actual number is changed most of the time the value is a DOUBLE but in some paths it is an INTEGER.  The "fix" is to handle both cases.

If this is acceptable then I will create the hotfix PR.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
@@ -648,11 +648,17 @@ public class WardPharmacyRectify extends JDialog {
 	 */
 	private JSpinner getJSpinnerNewQty() {
 		if (jSpinnerNewQty == null) {
-			jSpinnerNewQty = new JSpinner(new SpinnerNumberModel(0.0, 0.0, null, 1));
+			jSpinnerNewQty = new JSpinner(new SpinnerNumberModel(0.0d, 0.0d, null, 1.0d));
 			jSpinnerNewQty.setFont(new Font("Tahoma", Font.BOLD, 14)); //$NON-NLS-1$
 			jSpinnerNewQty.addChangeListener(changeEvent -> {
 				Double stock = Double.parseDouble(jLabelStockQty.getText());
-				Double newQty = (Double) jSpinnerNewQty.getValue();
+				Double newQty;
+				Object qtyValue = jSpinnerNewQty.getValue();
+				if (qtyValue instanceof Double) {
+					newQty = (Double) qtyValue;
+				} else {
+					newQty = Double.valueOf((Integer)qtyValue);
+				}
 				if (stock > 0) {
 					jButtonChooseLot.setEnabled(true);
 				}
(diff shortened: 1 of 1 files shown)


mwithi (commented):
It looked strange, so I've investigated a bit. According with the javadoc ([here](https://docs.oracle.com/javase/7/docs/api/javax/swing/SpinnerNumberModel.html)) relying on the model instead of the component should do the trick:
```
SpinnerNumberModel model = new SpinnerNumberModel(0.0d, 0.0d, null, 0.5d);
jSpinnerNewQty = new JSpinner(model);
jSpinnerNewQty.setFont(new Font("Tahoma", Font.BOLD, 14)); //$NON-NLS-1$
jSpinnerNewQty.addChangeListener(changeEvent -> {
	Double stock = Double.parseDouble(jLabelStockQty.getText());
	Double newQty = (Double) model.getNumber().doubleValue();
```

dbmalkovsky:
OK; made your changes and a few more that I couldn't resist given I was in the file :)

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -366,19 +372,15 @@ private void initComponents() {
 			} catch (OHServiceException e2) {
 				OHServiceExceptionUtil.showMessages(e2);
 			}
-			double newQty = (Double) jSpinnerNewQty.getValue();
+			double newQty = (Double)spinnerNewQtyModel.getNumber().doubleValue();
```
mwithi, line 375:
```suggestion
			double newQty = (Double) spinnerNewQtyModel.getNumber().doubleValue();
```

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -669,12 +672,15 @@ private JSpinner getJSpinnerNewQty() {
...
 			jLabelStockQty.setHorizontalAlignment(SwingConstants.CENTER);
 			jLabelStockQty.setPreferredSize(new Dimension(100, 25));
-			jLabelStockQty.setFont(new Font("Tahoma", Font.BOLD, 14)); //$NON-NLS-1$
+			jLabelStockQty.setFont(FONT_BOLD);
+			jLabelStockQty.setForeground(Color.BLACK);
+			jLabelStockQty.setEditable(false);
+			jLabelStockQty.setBorder(BorderFactory.createEmptyBorder());
 		}
 		return jLabelStockQty;
 	}
```
mwithi, line 686: should the variable also be renamed to jTextFieldStockQty? 

Anyway I was suggesting to not "parse" its text, rather get it directly from the source variable

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
dbmalkovsky: OK; added a space

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
dbmalkovsky: OK; made the name change.   Sorry I didn't understand your original comment; but given that it is a text field this just feels better.

mwithi (approved):
LGTM. Thanks!

PR #1267, opened 2022-08-02
OP-949 missed two renames

dbmalkovsky:
See OP-949

While creating the hotfix version I noticed that I forgot to rename two other fields/metho names in the original PR; sorry


mwithi (approved):
Alright! Thanks a lot!

PR #1270, opened 2022-08-08
OP-917 fix rendering of Rectify button

dbmalkovsky:
See OP-917  this addresses only the Pharmacy → Pharmaceutical Stock Ward → Drugs [tab] after the New button is used.

This is a bit heavy handed but after trying multiple options of reevaluate, invalidate, repaint, resetting the column widths, etc. nothing else worked.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
@@ -1486,8 +1486,6 @@ public class WardPharmacy extends ModalJFrame implements
 					WardPharmacyRectify wardRectify = new WardPharmacyRectify(WardPharmacy.this, wardSelected, medic);
 					wardRectify.addMovementWardListener(WardPharmacy.this);
 					wardRectify.setVisible(true);
-					TableCellRenderer buttonRenderer = new JTableButtonRenderer();
-					jTableDrugs.getColumn("").setCellRenderer(buttonRenderer);
 				});
 				return button;
 			}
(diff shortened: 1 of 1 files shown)


mwithi:
I found something concerning that button that is strange, but sure could be optimized:
lines 516/517 and 1490/1491 and 1535/1536 are redundant...

maybe setting the viewport is the one rendering better the cellrenderers,
but maybe setting a cellrenderer again and again in other methods causes the glitch.

dbmalkovsky:
I tried various combinations of commenting out the assignment of cell renderers and the problem still exists.   Guess my heavy handed approach is the one to go with.  Sorry.

mwithi:
> I tried various combinations of commenting out the assignment of cell renderers and the problem still exists. Guess my heavy handed approach is the one to go with. Sorry.

There is another chance maybe: `jTableDrugs.setAutoCreateColumnsFromModel(false)` when we create the table, in this way should be enought to set the rendereres only once for all.

dbmalkovsky:
@mwithi Nice find on `jTableDrugs.setAutoCreateColumnsFromModel(false)` is seems to work.  Thanks.

mwithi:
> @mwithi Nice find on `jTableDrugs.setAutoCreateColumnsFromModel(false)` is seems to work. Thanks.

Do you think we can try also with removing the duplicated code above?

dbmalkovsky:
The duplicated code is removed and things appear to function correctly.

mwithi:
> The duplicated code is removed and things appear to function correctly.

wonderful! thanks a lot!

mwithi (approved):
LGTM. Thanks!

PR #1271, opened 2022-08-08
OP-902 use STRONGLENGTH and validate password length

dbmalkovsky:
See OP-902

Password validation length is done in the GUI because by the time it gets to the IO level it is encrypted and a fixed length of 60 characters.


src/main/java/org/isf/menu/gui/UserEdit.java
```diff
@@ -238,10 +239,20 @@ private JButton getOkButton() {
...
+					if (password.length < GeneralData.STRONGLENGTH) {
+						MessageDialog.error(null, "angal.userbrowser.passwordmustbeatleastncharacters.fmt.msg", GeneralData.STRONGLENGTH);
+						return;
+					}
 					if (!Arrays.equals(password, repeatPassword)) {
 						MessageDialog.error(null, "angal.userbrowser.passwordsdonotmatchpleasecorrect.msg");
 						return;
 					}
+					// BCrypt has a maximum length of 72 characters
+					// see for example, https://security.stackexchange.com/questions/152430/what-maximum-password-length-to-choose-when-using-bcrypt
```
mwithi, line 251: 🔥😎 thanks!

mwithi (approved):
ALGTM and tested! Thanks a lot!

PR #1273, opened 2022-08-11
OP-954 update month and year choosers

dbmalkovsky:
See OP-954


mwithi (approved):
LGTM and tested also. Thanks!

PR #1274, opened 2022-08-12
OP-955 remove unused code

dbmalkovsky:
See OP-955


mwithi (approved):
LTGM and tested also. Thanks!

PR #1279, opened 2022-08-23
OP-958 remove CustomJDateChooser widget

dbmalkovsky:
See OP-958


mwithi (approved):
LGTM and tested. Thanks!

PR #1280, opened 2022-08-26
Chore: combine all recent dependabot jar updates

dbmalkovsky:
Will rebase dependabot PRs when this is committed.

Was tested with `mvn clean test` with the updated jar from https://github.com/informatici/openhospital-core/pull/734


mwithi (approved):
Same as https://github.com/informatici/openhospital-core/pull/734

PR #1281, opened 2022-08-29
OP-956 update the information in memory after a save

dbmalkovsky:
See OP-956

Also had to fix the refresh of the dialog when OPDEXTENDED=no


mwithi (approved):
LGTM. Thanks!

mwithi:
Seen latest change... the point is that we have to read the new lock value after the update, not between the read and the update, otherwise we are kind of tricking the JPA versioning.

dbmalkovsky:
Closing as a much better fix is found in https://github.com/informatici/openhospital-gui/pull/1287

PR #1286, opened 2022-09-02
OP-701-user-logout-and-automatic-logout (WIP)

goto-eof:
See OP-701
paired with https://github.com/informatici/openhospital-core/pull/738


src/main/java/org/isf/menu/gui/SplashWindow3.java
```diff
@@ -85,5 +91,13 @@ public void mousePressed(MouseEvent e) {
...
         splashThread.start();
     }
+    
+    private void startLogoutTimer(MainMenu mainMenu) {
+		if (UserSession.getTimer() != null) {
+			UserSession.getTimer().quit();
+		}
+		UserSession.setTimer(new DelayTimer(new LogoutEventListener(), GeneralData.SESSIONTIMEOUT));
+		UserSession.getTimer().startTimer();
+	}
```
dbmalkovsky, line 101: the formatting here looks wrong

src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -303,6 +307,10 @@ public void windowClosing(WindowEvent e) {
 			}
 		});
 		
+		if (!singleUser) {
+	           UserSession.setMainMenu(this);
+			}
+
```
dbmalkovsky, line 313: formatting?

src/main/java/org/isf/menu/gui/LogoutEventListener.java
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+	}
+
+	@Override
+	public void trigger() {
+		if (!GeneralData.getGeneralData().getSINGLEUSER() && UserSession.isLoggedIn()) {
+				UserSession.restartSession();
+				UserSession.getTimer().startTimer();
+				System.out.println("RESTARTED...");
```
dbmalkovsky, line 73: Shouldn't this be a logger instead of a `println`?

src/main/java/org/isf/menu/gui/LogoutEventListener.java
```diff
@@ -0,0 +1,79 @@
...
+import org.isf.generaldata.GeneralData;
+import org.isf.menu.model.User;
+import org.isf.utils.jobjects.DelayTimerCallback;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LogoutEventListener implements AWTEventListener, Login.LoginListener, DelayTimerCallback {
+
+	@Override
+	public void eventDispatched(AWTEvent e) {
```
dbmalkovsky, line 39: This seems to be very resource intensive.   Would it be better to do something with the database connection?  Almost every window/screen/dialog hits the database at some point.  Granted it would not be as fine grained as this but it would be a good approximation.    Just a thought ....

AndreiDodu: I agree with you that this solution consumes a lot of resources. Initially I restarted the timer when mouse_enter was done on some element of the screen. It would be necessary to leave the reset timer feature when the user uses the keyboard. 
I am wondering if Open Hospital has pages on which the user could not enter data or move the mouse.

src/main/java/org/isf/session/LoginEventListener.java
```diff
@@ -0,0 +1,20 @@
+package org.isf.session;
```
dbmalkovsky, line 1: Missing license header

src/main/java/org/isf/menu/gui/Login.java
```diff
@@ -122,6 +124,29 @@ public void keyReleased(KeyEvent event) {
...
+		Toolkit kit = Toolkit.getDefaultToolkit();
+		Dimension screensize = kit.getScreenSize();
+
+		Dimension mySize = getSize();
+
+		setLocation((screensize.width - mySize.width) / 2,
+				(screensize.height - mySize.height) / 2);
+
+		setResizable(false);
+		setVisible(true);
```
dbmalkovsky, line 147: There has been an effort to remove the user of Toolkit, screensize, setLocation, etc. and replace it with something like:
<pre>
         	pack();
		setLocationRelativeTo(null);
		setResizable(false);
		setVisible(true);
</pre>

mwithi:
Superseded by https://github.com/informatici/openhospital-gui/pull/1347.

PR #1287, opened 2022-09-03
OP-956 Fix OPD edit from Admission-Data-OPD-EDIT

mwithi:
See OP-956.

This should supersed https://github.com/dbmalkovsky/openhospital-gui/tree/OP-956.

Cc: @dbmalkovsky


dbmalkovsky (approved):
Looks very good to me and all my testing seems correct.   Thanks!

PR #1290, opened 2022-09-05
OP-942 Better handling of multiple updates

dbmalkovsky:
See [OP-942](https://openhospital.atlassian.net/browse/OP-942)

After seeing the fix in https://github.com/informatici/openhospital-gui/pull/1287 and the related discussion I went back and found a better solution for this issue that better honors the lock variable.

After update immediately set the lock variable in the current in memory table to the value just changed.


mwithi (approved):
LGTM. Thanks!

PR #1307, opened 2022-09-22
OP-844 add DICOM button in Admission/Patient Browser

dbmalkovsky:
See [OP-844](https://openhospital.atlassian.net/browse/OP-844)

Note: the accelerator key had to change because of conflicts.   

Note: the button name will change to Image when [OP-846](https://openhospital.atlassian.net/browse/OP-846) is done.

### DIFF
--- a/src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
+++ b/src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
@@ -725,12 +725,12 @@ public class AdmittedPatientBrowser extends ModalJFrame implements
 		if (MainMenu.checkUserGrants("data")) {
 			buttonPanel.add(getButtonData());
 		}
-		/*
-		 * Extra / experimental / temporary features - see Admin Manual
-		 */
 		if (GeneralData.DICOMMODULEENABLED && MainMenu.checkUserGrants("btnadmdicom")) {
 			buttonPanel.add(getDICOMButton());
 		}
+		/*
+		 * Extra / experimental / temporary features - see Admin Manual
+		 */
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/admission/gui/AdmittedPatientBrowser.java
```diff
@@ -725,12 +725,12 @@ private JPanel getButtonPanel() {
...
 		}
-		/*
-		 * Extra / experimental / temporary features - see Admin Manual
-		 */
 		if (GeneralData.DICOMMODULEENABLED && MainMenu.checkUserGrants("btnadmdicom")) {
 			buttonPanel.add(getDICOMButton());
 		}
+		/*
+		 * Extra / experimental / temporary features - see Admin Manual
+		 */
```
mwithi, line 733: Why moving this comment?

mwithi: (it can be just removed)

dbmalkovsky: the comment was above the extra/experimental features and this PR make the DICOM button no longer experimental but part of the product so I moved the comment below the adding button code

mwithi: misunderstanding... the comment was for the method itself, like the one for getButtonLab()) (OP-845)
so it can be removed

dbmalkovsky: done

mwithi (approved):
LGTM. Thanks!

PR #1312, opened 2022-09-26
OP-971 use TimeTools.truncateToSeconds() because it has a null check

dbmalkovsky:
See OP-971


mwithi (approved):
LGTM. Thanks!

mwithi (approved):
LGTM. Thanks!

PR #1313, opened 2022-09-26
OP-971 revert overzealous use of truncateToSeconds() method

dbmalkovsky:
See OP-971

found during interactive testing


mwithi (approved):
LGTM. Thanks!

PR #1314, opened 2022-09-27
OP-969 replace JPEGImageReader

dbmalkovsky:
See OP-969 

Uses `com.twelvemonkeys.imageio` module to get the code to compile with JDK 11


mwithi (approved):
Tested and it works fine! Thanks!

PR #1315, opened 2022-09-27
OP-781 drop LAB_EXAM_DATE field

dbmalkovsky:
See OP-774

Given that the field `LAB_EXAM_DATE` is removed by this Core PR: https://github.com/informatici/openhospital-core/pull/770 there were changes required in the JasperReports.

Before the change all 4 reports were generated and saved; after the change in CORE the same 4 reports were generated and compared manually and they all appeared to be the same.


rpt/OH007_lab_monthly_report.jrxml
```diff
@@ -22,23 +22,23 @@
    select EXA_DESC AS EXAM, LAB_MATERIAL AS MATERIAL, 'Tests done' AS CATEGORY, count(*) AS COUNT 
    from OH_LABORATORY, OH_EXAM
    where LAB_EXA_ID_A = EXA_ID_A
-   and month(LAB_EXAM_DATE) = $P{month} and year(LAB_EXAM_DATE) = $P{year}
+   and month(LAB_CREATED_DATE) = $P{month} and year(LAB_CREATED_DATE) = $P{year}
```
mwithi, line 25: the new field to rely on is `LAB_DATE`, `LAB_CREATED_DATE` is for audit purpose only

mwithi: same for all reports

dbmalkovsky: Change made in all reports

mwithi: Thanks!

src/main/java/org/isf/lab/gui/LabEditExtended.java
```diff
@@ -627,7 +627,6 @@ private JButton getOkButton() {
 				}
 				List<String> labRow = new ArrayList<>();
 				lab.setDate(TimeTools.getNow());
```
mwithi, line 629: this should be `lab.setDate(examDate);` as previously set from the `examDateFieldCal`

dbmalkovsky: Yes this is an issue that I'm pushing now

mwithi (approved):
LGTM. Thanks!!!

PR #1325, opened 2022-10-05
OP-977 allow "types" and User Group changes

dbmalkovsky:
See OP-977

Requires companion CORE PR https://github.com/informatici/openhospital-core/pull/780 to persist usergroup changes.

Also note added a check so a user cannot change their own assigned group.

Includes other changes for better variable names and other small code cleanup.


mwithi:
> Also note added a check so a user cannot change their own assigned group.

Very good! Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #1332, opened 2022-10-07
OP-875 link bill admission

mwithi:
See OP-875.
Check fails because they need https://github.com/informatici/openhospital-core/pull/782.


rpt/BillsReportMonthlyWard.properties
```diff
@@ -0,0 +1,21 @@
...
+
+Month = Month
+
+Receivable = Receivable
+
+Report.From = report from
+
+To = To
+
+Periodical.Incomes.Billed.And.Receivables.Report = Periodical Incomes/Billed and Receivables Report
```
dbmalkovsky, line 15:
```suggestion
Periodical.Incomes.Billed.And.Receivables.Report = Periodic Incomes/Billed and Receivables Report
```
The typical meaning of periodical is: a magazine or newspaper published at regular intervals.

rpt/BillsReportMonthlyWard_subreport.properties
```diff
@@ -0,0 +1,17 @@
...
+
+Month = Month
+
+Receivable = Receivable
+
+Report.From = report from
+
+To = To
+
+Periodical.Incomes.Billed.And.Receivables.Report = Periodical Incomes/Billed and Receivables Report
```
dbmalkovsky, line 15: same suggestion as above for `periodical`

dbmalkovsky (approved):
LGTM with small optional changes/suggestions

emecas (commented):
Looks good to me, @mwithi  in a previous PR you mentioned to me that one of the checks  was not relevant, here the following is the current situation, how could we rationalize it?

image

rpt/BillsReportMonthlyWard.properties
mwithi: Thanks!

mwithi: ok, but I need to change also the key, new commit comming!

mwithi:
> Looks good to me, @mwithi in a previous PR you mentioned to me that one of the checks was not relevant, here the following is the current situation, how could we rationalize it?

Because in previous PR the core was already merged (my bad) so it was passing only `pull_request` check, but normally both fails in first sight. Once the core is merged `pull_request` update itself automatically, `push` check must be relaunched manually.

PR #1334, opened 2022-10-07
 OP-780 - OH GUI make spring use xxx ACT (rebased) 

goto-eof:
See OP-780.
Paired with https://github.com/informatici/openhospital-core/pull/767.

### DIFF
--- a/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
+++ b/src/main/java/org/isf/admission/gui/AdmissionBrowser.java
@@ -1826,7 +1826,7 @@ public class AdmissionBrowser extends ModalJFrame {
 					admission.setType("N");
 				}
 
-				admission.setDeleted("N");
+				admission.setActive(1);
 
 				// IOoperation result
 				boolean result = false;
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/admission/gui/AdmissionBrowser.java
```diff
@@ -1830,7 +1830,7 @@ private JButton getSaveButton() {
 					admission.setType("N");
 				}
 
-				admission.setDeleted("N");
+				admission.setActive(1);
```
mwithi, line 1833: Is this command really needed now?

mwithi (changes requested):
The PR will not compile after the core merge because more methods have been renamed

src/main/java/org/isf/admission/gui/AdmissionBrowser.java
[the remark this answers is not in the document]
goto-eof: done

src/main/java/org/isf/supplier/gui/SupplierBrowser.java
```diff
@@ -337,7 +337,7 @@ public Object getValueAt(int r, int c) {
 			} else if (c == 7) {
 				return sup.getSupNote();
 			} else if (c == 8) {
-				return sup.getSupDeleted().equals('Y');
+				return sup.getActive().equals(0);
```
dbmalkovsky, line 340: Doesn't `getActive()` return an `int`?   So why isn't the test `return sup.getActive() == 1`?

PR #1335, opened 2022-10-07
OP-854 Move to JDK 11 - gui

dbmalkovsky:
See OP-854


emecas:
Just curious why this is failing, I saw the build job in Actions is set up for Java 8

mwithi:
> Just curious why this is failing, I saw the build job in Actions is set up for Java 8

yes, it needs to be removed, but only admin can change .github/workflows

dbmalkovsky:
@emecas it failed because "Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project OH-gui: Fatal error compiling: invalid flag: --release -> [Help 1]".

That is JDK 8 did not recognize the `--release` flag in the `pom.xml` that was added with JDK 9

PR #1338, opened 2022-10-07
OP-983 fix index out of bounds when filtering the list

dbmalkovsky:
See OP-983

Needed to recreate the `sorter` object and reattach it to the `table` when a change was made


mwithi (commented):
LGTM. Thanks a lot!

PR #1339, opened 2022-10-08
OP-983 remove the problematic sorting code

dbmalkovsky:
See OP-983

Made a few more changes to better match other settings browser's code


mwithi (approved):
LGTM. Thanks a lot!

PR #1340, opened 2022-10-08
OP-986 fix getting Ward column when other wards present

dbmalkovsky:
See OP-986 which includes how to reproduce the issue


mwithi (commented):
LGTM. Thanks a lot!

PR #1342, opened 2022-10-10
OP-875 link bill admission - gui

mwithi:
See OP-875
Checks fail because it needs https://github.com/informatici/openhospital-core/pull/786.


src/main/java/org/isf/accounting/gui/PatientBillEdit.java
```diff
@@ -112,6 +112,8 @@
 public class PatientBillEdit extends JDialog implements SelectionListener {
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(PatientBillEdit.class);
+	private static final ImageIcon admissionIcon = new ImageIcon("rsc/icons/bed_icon.png");
+	private static final String textOPD = MessageBundle.getMessage("angal.common.opd.txt");
```
dbmalkovsky, line 116: The variable names do not follow the standards for `static final` variables; see `LOGGER`

src/main/java/org/isf/accounting/gui/PatientBillEdit.java
```diff
@@ -386,7 +401,14 @@ private void initCurrencyCod() {
...
 
 	private void setBill(Bill bill) {
-		this.thisBill = bill;
+		try {
+			this.thisBill = (Bill) bill.clone();
+			this.thisBill.setAdmission(bill.getAdmission());
+			this.thisBill.setBillPatient(bill.getBillPatient());
+			this.thisBill.setPriceList(bill.getPriceList());
+		} catch (CloneNotSupportedException cnse) {
+			cnse.printStackTrace();
```
dbmalkovsky, line 410: Why isn't this Logged instead of printing a stack trace?

dbmalkovsky (approved):
made various small suggestions

src/main/java/org/isf/accounting/gui/PatientBillEdit.java
mwithi: It is a very rare case, I think it can be tracked for more info

emecas: you could add the full stack trace with all info to the log as well if that is the case, no sure if keeping the printStackTrace could add advantages for GUI debugging

emecas (approved):

PR #1343, opened 2022-10-11
OP-988 ensure result is set for procedure type 1

dbmalkovsky:
See OP-988

Note: could not use `indexOf()` to find the row because values could be null and the `equals()` in `Laboratory` doesn't handle null cleanly.


src/main/java/org/isf/lab/gui/LabNew.java
```diff
@@ -763,6 +770,7 @@ private JButton getJButtonAddExam() {
 					}
 
 					if (exa.getProcedure() == 1 || exa.getProcedure() == 3) {
+						// Typically procedure type 1 doesn't have a default value
```
mwithi, line 773:
```suggestion
						// The exam may not have a default value
```

PR #1347, opened 2022-10-14
OP-701 - user logout and automatic logout

goto-eof:
See OP-701.
Checks fail because it neeeds https://github.com/informatici/openhospital-core/pull/738.


src/main/java/org/isf/session/LogoutEventListener.java
```diff
@@ -0,0 +1,54 @@
...
+import java.awt.event.KeyEvent;
+
+import org.isf.generaldata.GeneralData;
+import org.isf.utils.jobjects.DelayTimerCallback;
+
+public class LogoutEventListener implements DelayTimerCallback, AWTEventListener {
+
+	@Override
+	public void trigger() {
+		boolean isNotSingleUserMode = !GeneralData.getGeneralData().getSINGLEUSER();
```
dbmalkovsky, line 35: What is single user mode even tested for; looking at the code the `LogoutEventListener` is only active in not in single user mode:
<pre>
			if (!GeneralData.getGeneralData().getSINGLEUSER()) {
				startLogoutTimer(mainMenu);
			}
</pre>

src/main/java/org/isf/session/LogoutEventListener.java
```diff
@@ -0,0 +1,54 @@
...
+		boolean loggedIn = UserSession.isLoggedIn();
+		if (isNotSingleUserMode && loggedIn) {
+			UserSession.restartSession();
+		}
+	}
+
+	@Override
+	public void eventDispatched(AWTEvent e) {
+		boolean isNotSingleUserMode = !GeneralData.getGeneralData().getSINGLEUSER();
```
dbmalkovsky, line 44: See line 35 above

src/main/java/org/isf/session/LogoutEventListener.java
```diff
@@ -0,0 +1,54 @@
...
+		boolean isLoggedIn = UserSession.isLoggedIn();
+		boolean isKeyEvent = e instanceof KeyEvent;
+		boolean isKeyPressed = KeyEvent.KEY_PRESSED == ((KeyEvent) e).getID();
+		boolean isTimerAvailable = UserSession.getTimer() != null;
+		if (isNotSingleUserMode && isLoggedIn && isKeyEvent && isKeyPressed && isTimerAvailable) {
```
dbmalkovsky, line 49: Why get all 5 booleans and then test them together?   Wouldn't it be more efficient to get and test the most unlikely one first and exit if not true?  Then get and test the next most unlikely, test it, and continue?

bundle/language_en.properties
```diff
@@ -891,6 +891,7 @@ angal.menu.btn.examtype
 angal.menu.btn.generaldata                                                                             = Settings
 angal.menu.btn.groups                                                                                  = Groups
 angal.menu.btn.help                                                                                    = Help
+angal.menu.btn.logout																				   = Logout
```
mwithi, line 894: fix alignment please

src/main/java/org/isf/menu/gui/Login.java
```diff
@@ -122,6 +124,29 @@ public void keyReleased(KeyEvent event) {
...
+		Toolkit kit = Toolkit.getDefaultToolkit();
+		Dimension screensize = kit.getScreenSize();
+
+		Dimension mySize = getSize();
+
+		setLocation((screensize.width - mySize.width) / 2,
+				(screensize.height - mySize.height) / 2);
+
+		setResizable(false);
+		setVisible(true);
```
mwithi, line 147: I report comment of @dbmalkovsky from previous PR... 

> There has been an effort to remove the user of Toolkit, screensize, setLocation, etc. and replace it with something like:
>
```
         	pack();
		setLocationRelativeTo(null);
		setResizable(false);
		setVisible(true);
```

src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -190,8 +195,7 @@ public MainMenu() {
 				}
 				new CommunicationFrame();
 				/*
-				 * Interaction communication= new Interaction();
-				 * communication.incomingChat(); communication.receiveFile();
+				 * Interaction communication= new Interaction(); communication.incomingChat(); communication.receiveFile();
```
mwithi, line 198:
```suggestion
				 * Interaction communication= new Interaction(); 
				 * communication.incomingChat(); communication.receiveFile();
```

bundle/language_en.properties
goto-eof: I don't see a wrong alignment

mwithi: not?
image

src/main/java/org/isf/accounting/gui/PatientBillEdit.java
```diff
@@ -680,9 +680,7 @@ private JPanel getJPanelDate() {
 			jPanelDate.add(getJCalendarDate());
 			jPanelDate.add(getJButtonPickPatient());
 			jPanelDate.add(getJButtonTrashPatient());
-			if (!GeneralData.getGeneralData().getSINGLEUSER()) {
-				jPanelDate.add(getJLabelUser());
-			}
+			jPanelDate.add(getJLabelUser());
```
mwithi, line 683: Why this change was needed?

src/main/java/org/isf/admission/gui/AdmissionBrowser.java
```diff
@@ -1826,8 +1826,6 @@ private JButton getSaveButton() {
 					admission.setType("N");
 				}
 
-				admission.setDeleted("N");
-
```
mwithi, line 1830: maybe this change comes from issue OP-780 that is still in review, so this piece of code is still needed

src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -386,13 +393,16 @@ public MainPanel(MainMenu parentFrame) {
 				if (u.getMySubmenu().equals("main")) {
 					button[k] = new JButton(u.getButtonLabel());
 					button[k].setMnemonic(KeyEvent.VK_A + (u.getShortcut() - 'A'));
-
-					button[k].addActionListener(parentFrame);
+					if ("logout".equals(u.getCode())) {
+						addLogoutButtonListener(k);
+					} else {
+						button[k].addActionListener(parentFrame);
+					}
```
mwithi, line 400: I don't know if it is really needed a MENUITEM for the logout: the button should always appear when SINGLEUSER=no, and should not appear when SINGLEUSER=yes.
Having a MENUITEM entry make sense when we want eventually disable the logout function for some user, but we do?
Of course, having this entry simplify the button construction in the main menu...

bundle/language_en.properties
goto-eof: I don't see it on linux. I replaced tabs with spaces.

mwithi (approved):
LGTM. Thanks a lot!

PR #1356, opened 2022-10-22
Chore: remove test as LocalTime object must be in range of 0 to 24

dbmalkovsky:
Because `startTime` and `endTime` are `LocalTime` objects the hour value ***must*** be in the range of 0 to 24 inclusive thus this test is not needed.


mwithi (approved):
LGTM. Thanks a lot!

PR #1357, opened 2022-10-22
OP-996 use correct variable in updateMedial() call

dbmalkovsky:
See OP-996

Also moved definition of `newMedical` to make the scope of the variable more clear.


mwithi (approved):
LGTM. Thanks a lot!

PR #1485, opened 2022-11-22
OP-347 Cumulative script update for 1.12

mizzioisf:
Batch of updates for the incoming 1.12 release.

Last update: clarified database server console messages
Update: added check for SERVER feature

New features:
- Full 64bit support on Windows, also for DICOM ! (no more ugly DICOM workarounds - see https://openhospital.atlassian.net/browse/OP-706)
- SERVER mode support (for portable installations) -> give it a try !
- Arabic Language support - oh.sh -l ar
- JRE 11 !
- MariaDB 10.6.x !

Along with the usual bug fixes, cosmetic/syntax changes, etc.

### DIFF
--- a/oh.ps1
+++ b/oh.ps1
@@ -173,7 +173,7 @@ $script:languagearray= @("en","fr","it","es","pt","ar")
 $script:MYSQL_VERSION="10.6.11"
 $script:MYSQL32_VERSION="10.6.5"
 
-######## define system architecture
+######## define system and software architecture
 $script:ARCH=$env:PROCESSOR_ARCHITECTURE
 
 $32archarray=@("386","486","586","686","x86","i86pc")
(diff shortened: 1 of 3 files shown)


oh.ps1
```diff
@@ -195,15 +196,18 @@ else {
...
-	$script:JAVA_PACKAGE_ARCH="i686";
-	#$script:MYSQL_ARCH=32;
-}
+#
+# NOT NEEDED ANYMORE FROM OH 1.12-dev !
+#
 
-######## MySQL/MariaDB Software
-# MariaDB
+#if ( $DICOM_ENABLE -eq "on" ) {
```
dbmalkovsky, line 203: Is this still needed?  I thought when upgraded/updated the DICOM jars and native libs this wasn't needed any longer.

dbmalkovsky: ugh ... I didn't notice the comment characters at the beginning of the line 😦

oh.ps1
```diff
@@ -523,6 +534,12 @@ function initialize_database {
 }
 
 function start_database {
+	Write-Host "Checking if MySQL server is running..."
+	if ( ( Test-Path "$OH_PATH/$TMP_DIR/mysql.sock" ) -or ( Test-Path "$OH_PATH/$TMP_DIR/mysql.pid" ) ) {
+		Write-Host "MySQL server already running! Exiting."
```
dbmalkovsky, line 539: There seems to be a move away from saying "MySQL" and using "database" instead which I think is a good idea given MySQL and MariaDB;  so there are other places that should change too?  Agreed?

mwithi: A second pass will be done for the doc

PR #1504, opened 2022-12-08
OH Startup scripts update (interactive menu)

mizzioisf:
2023.01.18
- Update JRE version (smaller package also)
- Finally fixed the libncurses check on Linux !

2022.12.20 -> New big chunk of updates, still  **__Work_in_Progress__**:
- interactive menu redesign -> still needs reviews <-
- direct edit of current option file (setting.properties): should be really easier for user now
- reviewed console output
- removed dicom workarounds, finally !
- better error checking
- bugfixes as usual
- quite a lot of new bugs :-)
----

Some quite big updates to the OH startup script **oh.sh** and **oh.ps1**. Needs some testing, but it might be more user friendly in the end. Some more updates (Dec 13th -> I would close here the functional updates of the scripts for 1.12):

-> Added "**-m**" option to configure OH manually
-> Various navigation updates
-> Parametric console message handling (MariaDB/MySQL correctly shown)
-> Better error handling
-> Updates and fixes

Thank you @dbmalkovsky for the review, did you test the linux or windows version?

-> **Interactive menu**: it is possible to navigate through menu options
-> Set **default to MULTIUSER** environment, so login mask is presented at startup
-> bug fixes and clean-up

Note: oh.sh can be used in the "standard" way with command line options (e.g. ./oh.sh -P) or interactively (./oh.sh)


oh.ps1
```diff
@@ -810,211 +795,263 @@ cd "$OH_PATH" # workaround for hard coded paths
...
+			Write-Host ""
+			get_confirmation;
+			set_language;
+			initialize_dir_structure;
+			mysql_check;
+			# ask user for database root password
+			$script:DATABASE_ROOT_PW = Read-Host "Please insert the MySQL / MariaDB database root password (root@$DATABASE_SERVER) -> "
+			Write-Host "Installing the database....."
+			Write-Host ""
+			Write-Host " Database name -> $DATABASE_NAME"
```
dbmalkovsky, line 871: Small nitpick: Would it make sense to write out the just entered $DATABASE_ROOT_PW to ensure they really entered what they thought they entered?  So later if they want to access the DB as root they really know what the password is.

oh.ps1
```diff
@@ -810,211 +795,263 @@ cd "$OH_PATH" # workaround for hard coded paths
...
-#		default { Write-Host "Invalid option: $opt. Exiting."; exit 1; }
-#		-> default is set to start with OH 
+		###################################################
+		"v"	{ # show version
+	        	Write-Host "--------- Software version ---------"
+			Get-Content $OH_PATH\$OH_DIR\rsc\version.properties | Where-Object {$_.length -gt 0} | Where-Object {!$_.StartsWith("#")} | ForEach-Object {
+			$var = $_.Split('=',2).Trim()
+			New-Variable -Force -Scope Private -Name $var[0] -Value $var[1] 
+			}
+			Write-Host "Open Hospital version" $VER_MAJOR $VER_MINOR $VER_RELEASE
```
dbmalkovsky, line 976: To match the other line formatting
```suggestion
			Write-Host "Open Hospital version: $VER_MAJOR $VER_MINOR $VER_RELEASE"
```

oh.ps1
```diff
@@ -810,211 +795,263 @@ cd "$OH_PATH" # workaround for hard coded paths
...
+		###################################################
+		default { Write-Host "Invalid option: $opt."; 
+			Read-Host;
+			break;
+		}
+		}
+	Clear-Host;
 	}
+	# execute until quit is pressed or CLIENT/PORTABLE/SERVER mode is select (Z option)
+	until ( ($opt -ceq 'q') -Or ($opt -ceq 'Z') )
```
dbmalkovsky, line 1054: above in lines 1031 and 1036 the test is for both lower case and upper case Q; the *until* only tests for lowercase; shouldn't it be case insensitive?  (my shell scripting is rusty)

oh.ps1
mizzioisf: It's actually considered a bad practice: Even Powershell comments it out...

oh.ps1
mizzioisf: Corrected, thank you !

oh.ps1
mizzioisf: It's not you but the awful poweshell syntax, you are right !
Powershell the Case-Sensitive Comparison (-ceq) -> Correctet with -ieq, thank you !

oh.ps1
```diff
@@ -799,222 +811,302 @@ function clean_files {
...
+			Write-Host 				""
+                	$script:LOG_LEVEL=Read-Host		"Please select log level (INFO|DEBUG)"
+			Write-Host 				""
+			Write-Host 				"***** Database configuration *****"
+			Write-Host 				""
+			$script:DATABASE_SERVER=Read-Host	"Enter database server IP address [DATABASE_SERVER]"
+			$script:DATABASE_PORT=Read-Host		"Enter database server TCP port [DATABASE_PORT]"
+			$script:DATABASE_NAME=Read-Host		"Enter database database name [DATABASE_NAME]"
+			$script:DATABASE_USER=Read-Host		"Enter database user name [DATABASE_USER]"
+			$script:DATABASE_PASSWORD=Read-Host	"Enter database password [DATABASE_PASSWORD][DATABASE_PASSWORD]"
```
mwithi, line 941:
```suggestion
			$script:DATABASE_PASSWORD=Read-Host	"Enter database password [DATABASE_PASSWORD]"
```

mwithi (approved):
LGTM. Thanks a lot!

PR #1512, opened 2022-12-20
OP-274 use common date from/date to label properties

dbmalkovsky:
See OP-274

Also did some cleanup of one file with results of IntelliJ Analyze command.

There is a corresponding documentation PR to follow.


mwithi (approved):
LGTM. Thanks a lot!

PR #1526, opened 2023-01-19
OP-684 Provide option to Launch online version of doc if not found locally

dbmalkovsky:
See OP-684

Added a Yes/No dialog message if the doc file is not found locally.

### DIFF
--- a/src/main/java/org/isf/help/HelpViewer.java
+++ b/src/main/java/org/isf/help/HelpViewer.java
@@ -24,8 +24,11 @@ package org.isf.help;
 import java.awt.Desktop;
 import java.io.File;
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
 
 import javax.swing.JDialog;
+import javax.swing.JOptionPane;
 
 import org.isf.generaldata.GeneralData;
 import org.isf.utils.jobjects.MessageDialog;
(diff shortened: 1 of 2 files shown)


src/main/java/org/isf/help/HelpViewer.java
```diff
@@ -40,28 +43,36 @@ public class HelpViewer extends JDialog {
 	private static final long serialVersionUID = 1L;
 
 	private static final String MANUAL_PDF_FILE = "UserManual.pdf";
+	private static final String USER_MANUAL_URI = "https://www.open-hospital.org/user-manual";
```
mwithi, line 46: I think it could be a good idea to link to the corresponding version on GitHub, for instance:

https://github.com/informatici/openhospital-doc/blob/1.11.5/doc_user/UserManual.adoc

taking the `1.11.5` from `version.properties`

dbmalkovsky: I disagree with that approach; it looks less official and I'm afraid people will be confused by the GIithub banners and other info there:
image

dbmalkovsky: I would assume must hospital staff don't even now what Github.   @mizzioisf Do you have an opinion?

mwithi: I don't think users would be confused by GitHub because they would go straight to the content.
The point is not GitHub or official website (GitHub is official too btw) but to link to the correct version of the doc.

We can think to publish each version in HTML somewhere online, I do remember that asciidoc was chosen to be ready for [readthedocs](https://readthedocs.org/) but I was trying without success (it needs an index.rst in the project)

dbmalkovsky: I used the link you provided to get the screen shot that I added.   Note the header, banners, icons, etc.   It just doesn't render the content.  It renders the content within the GitHub application.

mwithi: Yes, I know it very well, but it is the only solution we have so far for versioned doc

dbmalkovsky: It shouldn't be that hard to provide versioned documentation on the website; I've done it before.   But in the meantime I've changed it to go to Github.

mwithi: Yes as a temporary solution is better than nothing, thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #1529, opened 2023-01-23
OP-1023 Fix NPE by using the correct variable name

dbmalkovsky:
See OP-1023

### DIFF
--- a/src/main/java/org/isf/admission/gui/PatientFolderBrowser.java
+++ b/src/main/java/org/isf/admission/gui/PatientFolderBrowser.java
@@ -890,7 +890,7 @@ public class PatientFolderBrowser extends ModalJFrame
 
 			try {
 				labList = labManager.getLaboratory(patient);
-				getOlderDate(labList, "examDate");
+				getOlderDate(labList, "createdDate");
 			} catch (OHServiceException e) {
 				labList = new ArrayList<>();
 				OHServiceExceptionUtil.showMessages(e);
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/admission/gui/PatientFolderBrowser.java
```diff
@@ -890,7 +890,7 @@ public LabBrowserModel() {
 
 			try {
 				labList = labManager.getLaboratory(patient);
-				getOlderDate(labList, "examDate");
+				getOlderDate(labList, "createdDate");
```
mwithi, line 893:
```suggestion
				getOlderDate(labList, "labDate");
```

dbmalkovsky: @mwithi `labDate` will only work if there is a new method is added to the `Laboratory` object.   Reflection can't find it because the `getter` method is called `getDate` and not `getLabDate`.

If one looks at the history of the `Laboratory` object you'll see that the old comment suggested using `createDate`:

<pre>
	/**
	 * use <code>getCreatedDate()</code> instead
	 * @deprecated 
	 */
	@Deprecated
	public LocalDate getExamDate() {
		return examDate;
	}
</pre>

mwithi: yes, this beacuse the createDate and examDate got time, while labDate did not.
now labDate has seconds and can be used

dbmalkovsky: Variable name changed here and method added in CORE repo [PR](https://github.com/informatici/openhospital-core/pull/856)

mwithi (approved):
LGTM. Thanks a lot!

PR #1661, opened 2023-01-26
Script updates

mizzioisf:
This should be - hopefully - the final iteration before 1.12 release.

New: Desktop shortcut creation with icon, working on Windows (tested on 11) and Linux (Mint 21.1)
New: Reads/writes settings from existing configuration files

- Simplified options and usage
- Reworked and reorganized code
- Fixed some bugs from the latest reviews

Needs testing :-)


mwithi:
We have to solve once for all the problem of PORTABLE vs CLIENT... let's go for .ini? I think it's the best thing...

mwithi:
PS: on my windows there is no icon, normal?
image

mizzioisf:
> PS: on my Windows there is no icon, normal?
> 
> image

did you get the icon file from the PR also ?

mizzioisf:
> We have to solve once for all the problem of PORTABLE vs CLIENT... let's go for .ini? I think it's the best thing...

It's there ! I set the mode in settings.properties directly.

mwithi:
> > We have to solve once for all the problem of PORTABLE vs CLIENT... let's go for .ini? I think it's the best thing...
> 
> It's there ! I set the mode in settings.properties directly.

you're right! it works!

mwithi (approved):
LGTM. Thanks a lot!

PR #1662, opened 2023-01-26
OP-576 - filter bill manager table upon cashier

goto-eof:
See OP-576.

### DIFF
--- a/src/main/java/org/isf/accounting/gui/BillDataLoader.java
+++ b/src/main/java/org/isf/accounting/gui/BillDataLoader.java
@@ -51,13 +51,13 @@ public class BillDataLoader {
 
 		switch (status) {
 			case "O":
-				tableArray = getPendingBills(status);
+				tableArray = getPendingBills(status, username);
 				break;
 			case "ALL":
 				tableArray = getAllBills(username);
 				break;
 			case "C":
-				tableArray = getClosedBills(status);
+				tableArray = getClosedBills(status, username);
 				break;
 		}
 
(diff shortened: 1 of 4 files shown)


mwithi (approved):
Tested and looks fine for me! 
Thanks a lot!

src/main/java/org/isf/accounting/gui/BillDataLoader.java
```diff
@@ -81,17 +84,26 @@ private List<Bill> mergeBillsFromPeriodAndFromPayments() {
...
+		}
+		return list;
 	}
 
-	private List<Bill> getPendingBills(String status) throws OHServiceException {
-		return patientParent != null ? billManager.getPendingBillsAffiliate(patientParent.getCode()) :
-				billPeriod.stream()
+	private List<Bill> getPendingBills(String status, String username) throws OHServiceException {
+		if (patientParent != null) return  billManager.getPendingBillsAffiliate(patientParent.getCode()) ; 
+			List<Bill> list = billPeriod.stream()
```
dbmalkovsky, line 99: Line 98 should really be
<pre>
if (patientParent != null) {
   return  billManager.getPendingBillsAffiliate(patientParent.getCode());
}
</pre>
 
And then the formatting starting with line 99 looks indented too much.

mwithi: Sorry, merged to fast, changes applied on develop directly. Thanks!

PR #1664, opened 2023-01-27
OP-779 Fixed two reports after removing fields ADM_OPE_ID_A, ADM_DATE_OP, ADM_RESOP

mwithi:
See OP-779.

Paired with https://github.com/informatici/openhospital-core/pull/864


dbmalkovsky:
Remove `angal.admission.pleaseinsertavalidvisitdate.msg` from the resource bundle file; the string was removed in the CORE PR.

PR #1665, opened 2023-01-27
OP-527 inconsistent column definition for XXX_DELETED fields

mwithi:
See OP-527.

Paired with https://github.com/informatici/openhospital-core/pull/865.


dbmalkovsky (approved):
LGTM

PR #1675, opened 2023-01-31
OP-999: Hide subreports from list

nlazzer1:
Reports without jTitle property are not listed (e.g. subreports)


mwithi (approved):
LGTM. Thanks a lot!

PR #1682, opened 2023-01-31
Latest script fixes for 1.12

mizzioisf:
Latest fixes.
Set ohdemo database name.
Reordered settings.properties.dist file with comments.


rsc/settings.properties.dist
```diff
@@ -1,50 +1,70 @@
...
-INTERNALVIEWER=yes
-DOC_DIR=OH_DOC_DIR
-MAINMENUALWAYSONTOP=no
-RECEIPTPRINTER=yes
-VIDEOMODULEENABLED=yes
-PATIENTVACCINEEXTENDED=yes
-ENHANCEDSEARCH=no
-XMPPMODULEENABLED=no
+###########################################
+# dicom / imaging settings
```
dbmalkovsky, line 53: FWIW,  dicom comments in other files do not include "imaging"; should they?

mizzioisf: Yes, I think so, we should correct them when we find occurrences,

rsc/settings.properties.dist
```diff
@@ -1,50 +1,70 @@
+###########################################
 # This file contains Open Hospital settings
+###########################################
+# external settings
 MODE=OH_MODE
+###########################################
```
mwithi, line 6: I would replace ############### with an empty line

rsc/settings.properties.dist
mwithi: this file is for users (other files are scripts), it could just say "imaging" since we changed in this way in the doc (and not the code) but since it contains DICOM* parameters, I would leave it as "crossover" temporary solution.

PR #1685, opened 2023-02-05
OP-1028 include test for missing hour value too

dbmalkovsky:
See OP-1028

Reproduce by blanking out the time widget for a new visit.


mwithi (approved):
LGTM. Thanks a lot!

PR #1689, opened 2023-02-06
Demo data usage script updates

mizzioisf:
Fix and updates for DEMO data usage.
It should be now possible to switch form DEMO data and normal portable data seamlessy


mwithi (approved):
LGTM. Thanks a lot!

PR #1699, opened 2023-03-06
Update license-header.txt

mizzioisf:
Fixed URL.


dbmalkovsky (commented):
LGTM

mizzioisf:
> LGTM
> 
> We really don't need to do it as http gets redirected to https smile

Disagree here :-) anyway it is in line with the other URL in the file :-) :-)

PR #1702, opened 2023-03-07
OP-1044 fix menu size

dbmalkovsky:
See OP-1044

I am not proud of the fix but as this is how it was done in the original import of the code I just leveraged it instead of having to rewrite the whole layout of the dialog.

### DIFF
--- a/src/main/java/org/isf/stat/reportlauncher/gui/ReportLauncher.java
+++ b/src/main/java/org/isf/stat/reportlauncher/gui/ReportLauncher.java
@@ -298,7 +298,7 @@ public class ReportLauncher extends ModalJFrame{
 
 			jMonthComboBox.setSelectedIndex(month - 1);
 
-			jYearLabel = new JLabel("                  " + MessageBundle.getMessage("angal.stat.year"));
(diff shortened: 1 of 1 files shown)


mwithi:
"A bit more" I can still see border overlap
image

src/main/java/org/isf/stat/reportlauncher/gui/ReportLauncher.java
```diff
@@ -298,7 +298,7 @@ private JPanel getJParameterSelectionPanel() {
 
 			jMonthComboBox.setSelectedIndex(month - 1);
 
-			jYearLabel = new JLabel("        " + MessageBundle.getMessage("angal.stat.year"));
+			jYearLabel = new JLabel("                  " + MessageBundle.getMessage("angal.stat.year"));
```
mwithi, line 301:
```suggestion
			jYearLabel = new JLabel("                    " + MessageBundle.getMessage("angal.stat.year"));
```

mwithi: But I would add also a comment somewhere `//TODO: to fix layout` 😆

mwithi (approved):
LGTM. Thanks a lot!

PR #1726, opened 2023-03-22
Chore: combine all recent dependabot jar updates to match changes in Core

dbmalkovsky:
This also fixes the two CVEs noted in https://github.com/informatici/openhospital-core/pull/917


mwithi (approved):
LGTM. Thanks a lot!

PR #1727, opened 2023-03-22
OP-1035 - Patient Consensus Management

goto-eof:
See OP-1035.

Paired with https://github.com/informatici/openhospital-core/pull/905.

image

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1324,10 +1324,10 @@ angal.patient.newpatient.title                                                  
 angal.patient.nextkin                                                                                  = Next of Kin
 angal.patient.no.btn                                                                                   = No
 angal.patient.note                                                                                     = Note
-angal.patient.consensus.consensus.txt                                                                  = Consensus
-angal.patient.consensus.administrative.txt                                                             = Administrative
-angal.patient.consensus.service.txt                                                                    = Service
-angal.patient.consensus.border                                                                         = Consensus
+angal.patient.consensus.consensus.txt                                                                  = Consent to treatment
+angal.patient.consensus.administrative.txt                                                             = Administrative consent
+angal.patient.consensus.service.txt                                                                    = Consent for additional services
+angal.patient.consensus.border                                                                         = Informed consent
 angal.patient.parenttogether                                                                           = Parents are together
 angal.patient.patientbrowser.title                                                                     = Patient Browser
 angal.patient.patientphoto                                                                             = Patient Photo
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1324,6 +1324,10 @@ angal.patient.newpatient.title                                                  
 angal.patient.nextkin                                                                                  = Next of Kin
 angal.patient.no.btn                                                                                   = No
 angal.patient.note                                                                                     = Note
+angal.patient.consensus.consensus.txt                                                                  = Consensus
+angal.patient.consensus.administrative.txt                                                             = Administrative
+angal.patient.consensus.service.txt                                                                    = Service
+angal.patient.consensus.border                                                                         = Consensus
 angal.patient.parenttogether                                                                           = Parents are together
 angal.patient.patientbrowser.title                                                                     = Patient Browser
 angal.patient.patientphoto                                                                             = Patient Photo
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1325,6 +1325,7 @@ angal.patient.nextkin                                                           
 angal.patient.no.btn                                                                                   = No
 angal.patient.note                                                                                     = Note
 angal.patient.consensus.consensus.txt                                                                  = Consent to treatment
+angal.patient.consensus.consensus.mandatory.txt                                                        = "Consent to treatment" is mandatory
 angal.patient.consensus.administrative.txt                                                             = Administrative consent
 angal.patient.consensus.service.txt                                                                    = Consent for additional services
 angal.patient.consensus.border                                                                         = Informed consent
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1325,7 +1325,7 @@ angal.patient.nextkin                                                           
 angal.patient.no.btn                                                                                   = No
 angal.patient.note                                                                                     = Note
 angal.patient.consensus.consensus.txt                                                                  = Consent to treatment
-angal.patient.consensus.consensus.mandatory.txt                                                        = "Consent to treatment" is mandatory
+angal.patient.consensus.consensus.mandatory.msg                                                        = "Consent to treatment" is mandatory.
 angal.patient.consensus.administrative.txt                                                             = Administrative consent
 angal.patient.consensus.service.txt                                                                    = Consent for additional services
 angal.patient.consensus.border                                                                         = Informed consent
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -2166,12 +2177,60 @@ private JPanel getJRightPanel() {
...
+		checkbox.addActionListener(new ActionListener() {
+			@Override
+			public void actionPerformed(ActionEvent e) {
+				consensus.setServiceFlag(!consensus.isServiceFlag());
+				}
+			});
+		checkbox.setSelected(consensus.isServiceFlag());
+		panel.add(checkbox);
+		panel
+		.setBorder(
```
dbmalkovsky, line 2223: looks like some strange formatting?

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -2166,12 +2177,60 @@ private JPanel getJRightPanel() {
...
+	private JPanel getConsensus() {
+		try {
+			consensus = this.patientConsensusManager.getPatientConsensusByUserId(patient.getCode()).get();
+		} catch (RuntimeException | OHServiceException e) {
+			consensus = new PatientConsensus(); 
+		}
+		JPanel panel = new JPanel();
+		panel.setLayout(new BoxLayout(panel, BoxLayout.Y_AXIS));
+		JCheckBox checkbox = new JCheckBox(MessageBundle.getMessage("angal.patient.consensus.consensus.txt"));
+		checkbox.addActionListener(new ActionListener() {
```
dbmalkovsky, line 2196: This most likely could be lambda expressions instead of overridden methods.

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -28,6 +28,7 @@
 import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Image;
+import java.awt.event.ActionEvent;
```
mwithi, line 31: This import seems no longer needed

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -547,7 +561,10 @@ private JButton getJOkButton() {
 						}
 					}
 				} else {// Update
-
+					if (!consensus.isConsensusFlag()) {
+						MessageDialog.error(null, "angal.patient.consensus.consensus.mandatory.msg");
+						return;
+					}
```
mwithi, line 567: copy this check in the `PatientBrowserManger` into `validatePatient()` method.
It can stay in the gui if you like, but it is not mandatory to have it because the save would be refused by the manager

bundle/language_en.properties
```diff
@@ -1327,6 +1327,11 @@ angal.patient.newpatient.title
 angal.patient.nextkin                                                                                  = Next of Kin
 angal.patient.no.btn                                                                                   = No
 angal.patient.note                                                                                     = Note
+angal.patient.consensus.consensus.txt                                                                  = Consent to treatment
+angal.patient.consensus.consensus.mandatory.msg                                                        = "Consent to treatment" is mandatory.
+angal.patient.consensus.administrative.txt                                                             = Administrative consent
+angal.patient.consensus.service.txt                                                                    = Consent for additional services
+angal.patient.consensus.border                                                                         = Informed consent
```
dbmalkovsky, line 1334:
```suggestion
angal.patient.consensus.border                                                                         = Informed Consent
```
The border labels have the first letter of each word as a capital letter.

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -82,31 +88,23 @@
...
- * 11/08/2008 - alessandro - added mother and father names textfield
- * 11/08/2008 - alessandro - changed economicStatut -> hasInsurance
- * 19/08/2008 - mex        - changed educational level with blood type
- * 26/08/2008 - cla		   - added calendar for calculating age
- * 						   - modified age field from int to varchar
- * 28/08/2008 - cla		   - added tooltip for age field and checking name and age for patient editing
- * 05/09/2008 - alex       - added patient code
- * 01/01/2009 - Fabrizio   - modified assignment to age field to set an int value
- * ------------------------------------------
+ * ------------------------------------------ PatientInsertExtended - model for the patient entry ----------------------------------------- modification history
```
dbmalkovsky, line 91: Please undo the formatting the revision comments

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -2160,11 +2183,51 @@ private JPanel getJRightPanel() {
...
+		checkbox.setSelected(consensus.isServiceFlag());
+		panel.add(checkbox);
+
+		checkbox = new JCheckBox(MessageBundle.getMessage("angal.patient.consensus.administrative.txt"));
+		checkbox.addActionListener(e -> consensus.setAdministrativeFlag(!consensus.isAdministrativeFlag()));
+		checkbox.setSelected(consensus.isAdministrativeFlag());
+		panel.add(checkbox);
+
+		panel.setBorder(BorderFactory.createCompoundBorder(
+						BorderFactory.createCompoundBorder(
```
dbmalkovsky, line 2222: is this correct? `createCompoundBorder()` within another `createCompoundBorder()`; doesn't this look odd?

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
mwithi: We have to agree that having devs to enable "auto-format-upon-save" IDE's feature, will make hard to avoid such cases... we should either avoid comments reformating or just remove this old way of tracking revisions, wdyt?

dbmalkovsky: I vote to remove the old way of tracking revisions; that is what `git history` is very good at doing

mwithi: Ok, I think we can open an issue for this, so we can handle also other cases, wdyt?

src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -417,11 +425,21 @@ private JButton getJAnamnesisButton() {
...
 
+	private void validateConsensus(PatientConsensus patientConsensus) throws OHDataValidationException {
+		List<OHExceptionMessage> errors = new ArrayList<>();
+		if (patientConsensus == null || !patient.getPatientConsensus().isConsensusFlag()) {
+			errors.add(new OHExceptionMessage(MessageBundle.getMessage("angal.patient.consensus.consensus.mandatory.msg")));
+		}
+		if (!errors.isEmpty()) {
+			throw new OHDataValidationException(errors);
+		}
+	}
```
mwithi, line 441: If the patient is a new patient this method generate a NPE. It should be moved to `patientConsensusManager` and so will be called after `patientBrowserManager.savePatient(patient);`

mwithi (changes requested):
Creation of a new patient generates a NPE. See comments

mwithi (approved):

PR #1731, opened 2023-03-26
OH-722 Validate that a visit does not overlap other visits

joel-costigliola:
related to https://github.com/informatici/openhospital-core/pull/910


mwithi (approved):
LGTM. Thanks a lot!

PR #1734, opened 2023-03-27
OP-717 DICOM BLOB in a separate table

VaheAvetikyan:
Paired PR for OP-717.
Fixed compilation errors in GUI component.


mwithi (approved):
LGTM. Thanks a lot!

PR #1771, opened 2023-05-17
OP-1079: setAlwaysOnTop(true) for Login dialog

dbmalkovsky:
See OP-1079

### DIFF
--- a/src/main/java/org/isf/menu/gui/Login.java
+++ b/src/main/java/org/isf/menu/gui/Login.java
@@ -148,6 +148,7 @@ public class Login extends JDialog implements ActionListener, KeyListener {
 		usersListLogin = GeneralData.getGeneralData().getUSERSLISTLOGIN();
 		
 		this.parent = parent;
+		this.setAlwaysOnTop(true);
 
 		addLoginListener(parent);
 


dbmalkovsky:
The way this code is setup changing the dialog into a frame was no small feat (at least for me).   I set up a hidden JFrame and launched the dialog from it.    It now appears when using ALT-TAB and in my menu bar:

image

mwithi:
I think this should be tested by @mizzioisf and other Linux users (new [develop release](https://github.com/informatici/openhospital/pull/202)? 🙂)

On windows it looks worst, now is like this:
image

Before was like this:
image

dbmalkovsky:
I am more than happy if someone else takes this over; I've spent too much time on it already trying to make the minimum number of changes to make it work better.  Sorry.

mwithi:
> I am more than happy if someone else takes this over; I've spent too much time on it already trying to make the minimum number of changes to make it work better. Sorry. I think `alwaysOnTop(true)` is the answer; users can move the windows on their screen so they don't overlap if that is the big issue.

I think it's ok as well for Windows, just looks different but at least I have an icon at the bottom, that before was not there

mwithi (approved):
LGTM. Thanks a lot!

mizzioisf:
David, it works, nice ! Thank you !

We can merge in my opinion. Let me just point out that the normal OH window doesn't have a name (the login windows is correctly called "login")
See attached scrheenshot:

image

mwithi:
> OH window doesn't have a name

I would open another issue for that. Anyway, the point is that the Main Menu is not wide enough to show properly any title (we had same propblem with username). See attached screenshot with a test title "Open Hospital"

image

PR #1772, opened 2023-05-17
OP-1072: provide way to unlock user's account

dbmalkovsky:
See OP-1072

Requires:  https://github.com/informatici/openhospital-core/pull/993

If this is OK, I will update the doc


mwithi (approved):
Not sure if I like the question mark on a column header, I think it works also without... not big issue otherwise... :)

LGTM. Thanks a lot!

dbmalkovsky:
@mwithi I removed the question mark.

mwithi (approved):
LGTM. Thanks a lot!

PR #1777, opened 2023-05-27
OP-898: gui implementation of login idle

dbmalkovsky:
See OP-898

Requires: https://github.com/informatici/openhospital-core/pull/1002


mwithi (approved):
LGTM. Thanks a lot!

PR #1778, opened 2023-05-29
Chore: combine recent dependabot jar updates to match changes in Core

dbmalkovsky:
The surefire plugin in Core is already set to 3.1.0.

Clean rebuild and test.


mwithi (approved):
LGTM. Thanks a lot!

PR #1805, opened 2023-07-26
Chore: remove unneeded boxing

dbmalkovsky:
`Integer.valueOf()` returns an `Integer` where as `Integer.parseInt()` returns an `int`


mwithi (approved):
LGTM. Thanks a lot!

PR #1808, opened 2023-08-22
Chore: use formatMessage() instead of concat()

dbmalkovsky:
Code inspection recommended using `+` instead of `concat()` when constants are involved.  The `better` fix is just to use the capabilities of the `formatMessage()` method.


mwithi (approved):
LGTM. Thanks a lot!

PR #1809, opened 2023-08-23
Chore: remove import statement for class in same package

dbmalkovsky:
There is no need to specify an import for a class in the same package.


mwithi (approved):
LGTM. Thanks a lot!

PR #1810, opened 2023-08-23
Chore: use Integer.parseInt() for String to int conversion

dbmalkovsky:
It has the advantage of not needing boxing/unboxing of value.   `Integer.parseInt()` has used in the same file in other locations.


mwithi (approved):
LGTM. Thanks a lot!

PR #1812, opened 2023-08-26
Chore: misc code improvements

dbmalkovsky:
Remove unused variables, methods, assignments, close streams, immediately return value instead of assigning it to a local variable, etc.


mwithi (approved):
LGTM. Thanks a lot!

PR #1813, opened 2023-08-26
Chore: JUnit improvements

dbmalkovsky:
JUnit5 is more tolerant regarding the visibilities of Test classes than JUnit4, which required everything to be public.
In this context, JUnit5 test classes can have any visibility but private, however, it is recommended to use the default package visibility, which improves readability of code.

Also used more typical method (or more `geeky`) calls in many cases; for example, used `.hasSize()`


mwithi (approved):
LGTM. Thanks a lot!

PR #1814, opened 2023-08-28
Chore: remove unused file

dbmalkovsky:
In a recent commit I changed the name of this file because I believed it wasn't used.   Given the strange name, the fact that nothing breaks, the file hasn't been updated in a long time and the version in CORE has much more information I believe that in fact it can be deleted.


mwithi:
Sounds reasonable, but let's test it a little bit more. Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #1819, opened 2023-09-08
Chore: provide init value for variable

dbmalkovsky:
The variable `pfrmHeight` is only used once in a `new Dimension(225, pfrmHeight)` statement.   It didn't have assigned value and zero (the default) seems to work;   I tested with multiple values for `pfrmHeight` and 100 seems reasonable.   It could stay the way it is but it just looks funny/wrong IMHO.

### DIFF
--- a/src/main/java/org/isf/lab/gui/LabBrowser.java
+++ b/src/main/java/org/isf/lab/gui/LabBrowser.java
@@ -105,7 +105,7 @@ public class LabBrowser extends ModalJFrame implements LabListener, LabEditListe
 	private JPanel jSelectionPanel;
 	private JTable jTable;
 	private JComboBox comboExams;
-	private int pfrmHeight;
+	private int pfrmHeight = 100;
 	private List<Laboratory> pLabs;
 	private String[] pColumns = {
 			MessageBundle.getMessage("angal.common.date.txt").toUpperCase(),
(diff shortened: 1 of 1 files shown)


mwithi:
> The variable `pfrmHeight` is only used once in a `new Dimension(225, pfrmHeight)` statement. It didn't have assigned value and zero (the default) seems to work; I tested with multiple values for `pfrmHeight` and 100 seems reasonable. It could stay the way it is but it just looks funny/wrong IMHO.

The point is that the constructor `new Dimension` does require both parameters, but each component dimension can be honored or ignored independently, according by the layout manager applied to its parent container, we can't do differently in order to set that width.

PS: I would also remove the full package name `java.awt.` when using constants

dbmalkovsky:
The point is that there is a variable `pfrmHeight` that is used in the `Dimension` statement  has **NO** value other than the default value of zero.   So then either delete the variable from the `Dimension` statement, put zero for the height and remove it from the class or then initialize the variable to zero (I chose something that was more reasonable given the layout of the window).   Having an "uninitialized" variable and it never being set  just seems wrong when reading the code and wastes space on the stack.

mwithi:
> The point is that there is a variable `pfrmHeight` that is used in the `Dimension` statement has **NO** value other than the default value of zero. So then either delete the variable from the `Dimension` statement, put zero for the height and remove it from the class or then initialize the variable to zero (I chose something that was more reasonable given the layout of the window). Having an "uninitialized" variable and it never being set just seems wrong when reading the code and wastes space on the stack.

I agree with you, the way it is in this PR looks a good compromise for code readability and maintenance. 

Are you going to remove the full package name java.awt. when using constants too?

dbmalkovsky:
> Are you going to remove the full package name java.awt. when using constants too?

Yes .. done

PR #1820, opened 2023-09-09
Chore: remove unused code

dbmalkovsky:
A previous PR moved the definition of `List<Ward> ward` from a class instance to a local variable.  With it being local it is now obvious that the variable `ward` is not used.   Looking at the git history, this piece of code has been this way since moving to github since 2015.   It is just not needed.


mwithi (approved):
LGTM. Thanks a lot!

PR #1822, opened 2023-09-12
Chore: remove unused variable and assignment

dbmalkovsky:
The variable `patientBrowserManager` is never used and the assignment to `self` is never used again


mwithi (approved):
LGTM. Thanks a lot!

PR #1823, opened 2023-09-12
Chore: remove unused code and methods

dbmalkovsky:
The variable, `medicalJTextField` was previously used in a panel but the panel was removed leaving this variable defined, set, and never used.

The methods, `getPatientParent` and `setPatientParent` were never accessed.


mwithi (approved):
LGTM. Thanks a lot!

PR #1824, opened 2023-09-13
Chore: remove unused BillItemPicker

dbmalkovsky:
The file `BillItemPicker` is not used and should be removed.

The fact that `BillItemPicker` is not used was determined as follows:

- Noticed that the routine `setParentFrame()` (near the bottom of the file) was never called by any code
- If `setParentFrame()` wasn't called then the variable `parentFrame` was never set and remained the default value of `null`
- If `parentFrame` was `null` then code like `this.getParentFrame().dispose();` (line 304 in the original file) should be throwing a NPE (which it wasn't)
- That led me to search for the string `BillItemPicker` in all the files the GUI repo and the only occurrence (other than a single line in the resource bundle) was the file itself
- Removed the file and recompiled cleanly
- Ran multiple interactive tests of the GUI and especially the accounting operations and didn't find any issues
- Finally searched for the string `BillItemPicker` in all the files on the Core repo with no hits
- Conclusion:  remove the file and the single resource bundle entry


mwithi:
Some code went lost... the BillItemPicker was introduced with OP-90 and used in BillBrowser class in 2019: it was supposed to filter bills by a specific product. Anyway, since nobody claimed that missing button, I think it was lost even before the release [1.9.0-beta1](https://github.com/informatici/openhospital/releases/tag/v1.9-pre) (when we migrated from SourceForge to GitHub) leaving that class orphan... In case someone will claim now that feature, at least we know what we are talking about 😅

dbmalkovsky:
Let's just remove it; if someone wants it (with all its problems) they can find it in git history.   There is no need to bloat the code base.

mwithi (approved):
LGTM. Thanks a lot!

PR #1825, opened 2023-09-13
Chore: remove unused nested class

dbmalkovsky:
The nested class `ClipMover` is never referenced; remove from the code base.


mwithi (approved):
LGTM. Thanks a lot!

PR #1828, opened 2023-09-15
Chore: code cleanup and remove unused methods

dbmalkovsky:
First the method, `getSelectedObject()` is never called from any code.   Removing that method means the private method, `getSelectedRow()` is never referenced.  Removing that method means the method, `setSelectedRow()` is not needed as there isn't a way to reference the private variable,  `selectedRow` except by name which isn't done.   

Also removed fully qualified package names in the body of the code.


mwithi (approved):
LGTM. Thanks a lot!

PR #1832, opened 2023-09-16
Chore: remove getting hospital time range

dbmalkovsky:
First noticed that the method `getHospitalHours()` only had local variables and wasn't actually returning the hours.  So removed the method and the calls to the method.   The range of hours is not handled in a single place in the `GoodDateTimeVisitChooser` object.   

Cleaned up the imports and now unused `hospitalBrowsingManager` variable.

### DIFF
--- a/src/main/java/org/isf/opd/gui/OpdEditExtended.java
+++ b/src/main/java/org/isf/opd/gui/OpdEditExtended.java
@@ -21,8 +21,6 @@
  */
 package org.isf.opd.gui;
 
-import static org.isf.hospital.model.Hospital.VISIT_END_TIME;
-import static org.isf.hospital.model.Hospital.VISIT_START_TIME;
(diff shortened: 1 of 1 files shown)


mwithi:
> First noticed that the method `getHospitalHours()` only had local variables and wasn't actually returning the hours. So removed the method and the calls to the method. The range of hours is not handled in a single place in the `GoodDateTimeVisitChooser` object.
> 
> Cleaned up the imports and now unused `hospitalBrowsingManager` variable.

you mean ".. range of hours is **now** handled in ..."?

dbmalkovsky:
Yes, *now* and not *not* :frowning_face:   The letters *W* and *T* are too close to each other on the keyboard.

mwithi (approved):
LGTM. Thanks a lot!

PR #1835, opened 2023-09-19
OP-1126: add helpers in Worksheet

dbmalkovsky:
See OP-1126

Added "\<select a ward\>" message to initial ward combo box instead of blank.  Changed the behavior slightly as the old behavior when the "blank" entry was selected again after selecting a specific ward, all visits for all wards would be displayed; now, there is another entry "\<all wards\>" that can do  the same thing and selecting "\<select a ward\>" returns the dialog back to its original empty state.    This seems more consistent and understood behavior instead of "\<select a ward\>" meaning something different the first time and when used later.

I will update the documentation after this is committed.

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1684,7 +1684,6 @@ angal.visit.addvisit1.btn                                                       
 angal.visit.addvisit1.btn.key                                                                          = A
 angal.visit.addvisit2.btn                                                                              = Add Visit
 angal.visit.addvisit2.btn.key                                                                          = V
-angal.visit.allwards.txt                                                                               = <all wards>
 angal.visit.arrowprevious.btn                                                                          = <-Prev.
 angal.visit.avisitcannotbescheduledforadatethatispast.msg                                              = A visit cannot be scheduled for a date that is past.
 angal.visit.changepatient                                                                              = Change Patient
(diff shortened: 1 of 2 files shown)


mwithi:
Looks like the "Print this day's visits" report doesn't work with the new `<all wardss>` option (while specific selection does)

image

image

dbmalkovsky:
FWIW, this never worked.   If you selected a ward, added a visit, and set the ward back to blank and tried to print it you got a NPE.   How about I just disable the printing buttons when "<all wards>" are selected?

mwithi:
> FWIW, this never worked. If you selected a ward, added a visit, and set the ward back to blank (which is the new "<all wards>") and tried to print it you got a NPE. How about I just disable the printing buttons when "" are selected?
> 
> Another option would be to cycle through each of the wards and print a separate report for each one.
> 
> To print a combined report looks like a lot of work in code I don't understand.

Alright, I was remembering differently.

My suggestion is to open a bug issue for this: the report can just be modified to work with multiple wards (using Jasper GROUP feature) and then print a single ward or many depending by the selection (it is possible to pass a list and expand it in the Jasper QUERY language). Also the GUI should be fixed in order to show also the ward on each row in the table.

At this time I would just remove the _<all wards>_ option. For the "" case, is it possible to remove the "" option from the list upon first selection? Make sense?

dbmalkovsky:
Pushed code that removes `<all wards>` and removes `<select a ward>` from the combo box once the first selection in the list is made.

mwithi:
> Pushed code that removes `<all wards>` and removes `<select a ward>` from the combo box once the first selection in the list is made.

Works great!

mwithi (approved):
LGTM. Thanks a lot!

PR #1837, opened 2023-09-21
Chore: remove unused assignments

dbmalkovsky:
Examples include removing assignments that are the default value, the variable is replaced with another value, or the variable is never used again.


mwithi (approved):
LGTM. Thanks a lot!

PR #1879, opened 2023-11-08
OP-1130 improve JPA methods - patconsensus package

dbmalkovsky:
See OP-1130

Matches changes in CORE PR https://github.com/informatici/openhospital-core/pull/1158

### DIFF
--- a/src/main/java/org/isf/patient/gui/PatientInsertExtended.java
+++ b/src/main/java/org/isf/patient/gui/PatientInsertExtended.java
@@ -548,7 +548,7 @@ public class PatientInsertExtended extends JDialog {
 							if (justSave) {
 								insert = false;
 								justSave = false;
-                                this.requestFocus();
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/patient/gui/PatientInsertExtended.java
```diff
@@ -548,7 +548,7 @@ private JButton getJOkButton() {
 							if (justSave) {
 								insert = false;
 								justSave = false;
-								PatientInsertExtended.this.requestFocus();
+                                this.requestFocus();
```
mwithi, line 551: looks like the indentation is wrong

dbmalkovsky: fixed; thanks

mwithi (approved):
LGTM. Thanks a lot!

PR #1881, opened 2023-11-09
OP-1130 improve JPA methods - patient package

dbmalkovsky:
See OP-1130

Matches changes in https://github.com/informatici/openhospital-core/pull/1159

### DIFF
--- a/src/main/java/org/isf/patient/gui/PatientSummary.java
+++ b/src/main/java/org/isf/patient/gui/PatientSummary.java
@@ -133,7 +133,7 @@ public class PatientSummary {
 		return dataPanel;
 	}
 
-	final int insetSize = 5;
+	static final int insetSize = 5;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/isf/patient/gui/PatientSummary.java
```diff
@@ -133,7 +133,7 @@ private JPanel getPatientKinAndTelephonePanel() {
 		return dataPanel;
 	}
 
-	final int insetSize = 5;
+	static final int insetSize = 5;
```
mwithi, line 136: should it be INSETSIZE?

dbmalkovsky: yes; changed; thanks

mwithi (approved):
LGTM. Thanks a lot!

PR #1882, opened 2023-11-11
OP-1130 improve JPA methods - dicom package

dbmalkovsky:
See OP-1130

Changes in `dicom` package

Matches changes in https://github.com/informatici/openhospital-core/pull/1160

Fix spelling


mwithi (approved):
LGTM. Thanks a lot!

PR #1894, opened 2023-11-27
OP-1124-variable-operations-all-never-has-a-value-assigned 

nlazzer1:
Fix unused variables


src/main/java/org/isf/operation/gui/OperationRowBase.java
```diff
@@ -154,7 +152,7 @@ public void keyTyped(KeyEvent e) {
 		searchOperationButton.setIcon(new ImageIcon("rsc/icons/zoom_r_button.png"));
 		searchOperationButton.addActionListener(new ActionListener() {
 
-			List<Operation> operationsOPD;
+			List<Operation> operationsOPD = new LinkedList<>();
```
dbmalkovsky, line 155: while this is "OK"; should this variable be defined within the `actionPerformed` method at line 158?

mwithi (approved):
LGTM. Thanks a lot!

PR #1901, opened 2023-12-15
OP-1159 Fix patient profile picture in reports when loaded from FS

mwithi:
See OP-1159.


src/main/java/org/isf/patient/gui/PatientSummary.java
```diff
@@ -160,10 +163,18 @@ private JPanel getPatientCard() {
...
+		PatientProfilePhoto photo = null;
+		try {
+			photo = this.patientBrowserManager.retrievePatientProfilePhoto(patient);
+		} catch (OHServiceException e) {
+			OHServiceExceptionUtil.showMessages(e);
+		}
+
+		if (photo != null && photo.getPhotoAsImage() != null) {
+			patientPhoto.setIcon(
+							new ImageIcon(ImageUtil.scaleImage(photo.getPhotoAsImage(), GeneralData.IMAGE_THUMBNAIL_MAX_WIDTH)));
```
dbmalkovsky, line 177: surprised this is on a new line

dbmalkovsky (approved):
LGTM given my limited knowledge of JasperReports.   I had one small comment that can be ignored.

src/main/java/org/isf/patient/gui/PatientSummary.java
mwithi: it is the formatter doing this, if you fix and save the same file, what is it doing in your IDE?

PR #1906, opened 2023-12-19
Chore: remove unused variable/code

dbmalkovsky:
Defining a local variable in a `try` block and then not using it has no value.

Looking at the class history, the variable `ward` was originally a class variable that was used to fill a column in the table.  With time the column was modified to include `OPD` and the class variable was no longer used.  In another PR class variables that were used only locally were made into local variables and that is how this particular local was born.

Removing this code has no effect.


mwithi (approved):
LGTM. Thanks a lot!

PR #1909, opened 2023-12-19
OP-1130 improve JPA methods - operation package

dbmalkovsky:
See OP-1130

More changes for the `operation` package.

This matches changes in CORE PR https://github.com/informatici/openhospital-core/pull/1200


mwithi (approved):
LGTM. Thanks a lot!

PR #1910, opened 2023-12-21
OH2-252 | Fix audit logging

ArnaudFonzam:
See OH2-252.

- [x] remove DelayTimer and DelayTimerCallback to core
- [x] rename UserSession to RestartUserSession. UserSession is defined now on core. RestartUserSession class extend UserSession
- [x] update the import in all class with use UserSession to RestartUserSession


src/main/java/org/isf/session/LogoutEventListener.java
mwithi, line 1: Missing changes ?
- LoginEventListener
- SessionRefreshTimerRunnable

mwithi: Also
- CursorManager
- WaitCursorEventQueue

dbmalkovsky (approved):
LGTM

PR #1916, opened 2023-12-26
Chore: use compiled pattern instead of regular expression that needs to be regenerated each time

dbmalkovsky:


mwithi (approved):
LGTM. Thanks a lot!

PR #1930, opened 2024-01-10
OP-1140 | Height and Weight should be non-zero

loique70:
Add verification to check if Height and Weight are  non-zero before saving.


src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1465,17 +1465,33 @@ public SwingActionSavePatientExamination() {
...
+			if (patex.getPex_weight() != 0 && patex.getPex_height() != 0) {
+				if (patex.getPex_weight() == ExaminationParameters.WEIGHT_MIN || patex.getPex_weight() == ExaminationParameters.WEIGHT_MAX || patex.getPex_height() == ExaminationParameters.HEIGHT_MIN || patex.getPex_height() == ExaminationParameters.HEIGHT_MAX) {
+					int response = JOptionPane.showConfirmDialog(null, MessageBundle.getMessage("angal.patient.examination.defaultvalue.msg"), "Confirm", JOptionPane.YES_NO_OPTION);
+					if (response == JOptionPane.YES_OPTION) {
+						savePatientExamaination();
+					}
+				} else {
+					savePatientExamaination();
+				}
+			}else{
```
dbmalkovsky, line 1477:
```suggestion
			} else {
```

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1465,17 +1465,33 @@ public SwingActionSavePatientExamination() {
 		@Override
 		public void actionPerformed(ActionEvent actionEvent) {
 
-			try {
-				examinationBrowserManager.saveOrUpdate(patex);
-				modified = false;
-			} catch (OHServiceException ohServiceException) {
-				MessageDialog.showExceptions(ohServiceException);
+			if (patex.getPex_weight() != 0 && patex.getPex_height() != 0) {
```
dbmalkovsky, line 1468: Instead of fetching `weight` and `height` potentially many times why not get them once and store them in a local variable?

bundle/language_en.properties
```diff
@@ -1304,6 +1304,8 @@ angal.patient.dead.btn
 angal.patient.deletepatient.fmt.msg                                                                    = Delete patient: {0}?
 angal.patient.doyouwanttodeletethepatientsphoto.msg                                                    = Do you want to delete the patient's photo?
 angal.patient.editpatient.title                                                                        = Edit Patient
+angal.patient.examination.nonzero.msg                                                                  = Height or Weight must not be zero
```
dbmalkovsky, line 1307:
```suggestion
angal.patient.examination.nonzero.msg                                                                  = Height or weight must not be zero.
```

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
loique70: Thanks for the suggestion. This will  improve code and keep it clean.

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1464,18 +1464,36 @@ public SwingActionSavePatientExamination() {
...
-				examinationBrowserManager.saveOrUpdate(patex);
-				modified = false;
-			} catch (OHServiceException ohServiceException) {
-				MessageDialog.showExceptions(ohServiceException);
+			double weight = patex.getPex_weight();
+			int height = patex.getPex_height();
+
+			if (weight != 0 && height != 0) {
+				if (weight == ExaminationParameters.WEIGHT_MIN || weight == ExaminationParameters.WEIGHT_MAX || height == ExaminationParameters.HEIGHT_MIN || height == ExaminationParameters.HEIGHT_MAX) {
+					int response = JOptionPane.showConfirmDialog(null, MessageBundle.getMessage("angal.patient.examination.defaultvalue.msg"), "Confirm", JOptionPane.YES_NO_OPTION);
```
mwithi, line 1472: I think that this message should be shown when `weight == ExaminationParameters.WEIGHT_INIT && weight == ExaminationParameters.WEIGHT_INIT` and we should change it to: "`Some values are equal to the default values. Do you really want to confirm them?`" in order to reuse it also for others `_INIT` parameters.

While for MIN and MAX we should shown a message like "`Some entered values are equal to their minimum or maximum values. Do you really want to confirm them?`" in order to reuse it also for others `_MIN` and `_MAX` parameters.

Wdyt?

dbmalkovsky: Why aren't you using `MessageDIalog.yesNo()` instead of `showConfirmDIalog()`?   Then you wouldn't have to worry about the fact that you didn't localize the `Confirm` string and this interaction would match all the other dialogs in the project.

src/main/java/org/isf/examination/gui/PatientExaminationEdit.java
```diff
@@ -1464,18 +1464,36 @@ public SwingActionSavePatientExamination() {
...
+					if (response == JOptionPane.YES_OPTION) {
+						savePatientExamaination();
+					}
+				} else {
+					savePatientExamaination();
+				}
+			} else {
+				JOptionPane.showMessageDialog(null, MessageBundle.getMessage("angal.patient.examination.nonzero.msg"), "Error", JOptionPane.ERROR_MESSAGE);
```
dbmalkovsky, line 1480: Here is another case where there is a helper method in the code base:  `MessageDialog.error()`   Once again it takes care of the localizing the string "Error" for you.

loique70:
Okay.

PR #1933, opened 2024-01-11
Chore: use lambda instead of anonymous method

dbmalkovsky:
Since JDK8

### DIFF
--- a/src/main/java/org/isf/telemetry/gui/TelemetryEdit.java
+++ b/src/main/java/org/isf/telemetry/gui/TelemetryEdit.java
@@ -365,8 +365,9 @@ public class TelemetryEdit extends ModalJFrame {
             	    TelemetryDaemon.getTelemetryDaemon().stop();
             	    dispose();
        	        }
-        };
-
+            };
+        }
+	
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/telemetry/gui/TelemetryEdit.java
```diff
@@ -337,50 +337,34 @@ private Map<String, Boolean> buildConsentData(List<CheckBoxWrapper> permissions)
...
 
 	private ActionListener buildAskMeLaterButtonActionListener(TelemetryManager telemetryManager) {
-		return new ActionListener() {
-
-			public void actionPerformed(ActionEvent e) {
-				dispose();
-			}
-		};
-
+		return e -> dispose();
```
mwithi, line 340: true, and if I recall well, we decided also to use extended name, like `actionEvent` in this case?

```suggestion
		return actionEvent -> dispose();
```

dbmalkovsky: @mwithi Done and pushed

dbmalkovsky:
Closing this PR; the manual merge of the conflict messed things up badly.

PR #1939, opened 2024-01-16
OP-1168 Add LogViewer

mwithi:
See OP-1168.

Paired with https://github.com/informatici/openhospital-core/pull/1241

### DIFF
--- a/src/main/java/org/isf/utils/log/LogViewer.java
+++ b/src/main/java/org/isf/utils/log/LogViewer.java
@@ -1,0 +1,52 @@
+/*
+ * Open Hospital (www.open-hospital.org)
+ * Copyright © 2006-2024 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ *
+ * Open Hospital is a free and open source software for healthcare data management.
+ *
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/utils/log/LogViewer.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.isf.utils.log;
```
dbmalkovsky, line 22: I don't think it is a good idea to use the same package name in two different jars.    Why isn't this class at the same level as `org.isf.help`?  They both get invoked from the same menu and show a "window".

mwithi: There were many many reflections and attempts before this, but then an external POV is decisive, thanks!

PR #1946, opened 2024-01-26
OP-1174 update maven.yml for JDK 17

dbmalkovsky:
Matches https://github.com/informatici/openhospital-core/pull/1250


mwithi (approved):
LGTM. Thanks a lot!

PR #1950, opened 2024-01-31
Chore: small refactoring

mwithi:
Discovered while testing OP-782.

- [x] Refactor constants instead to be read over and over from memory while browsing the jtable
- [x] Refactor a dead code
- [x] Remove not used variable `wardBrowserManager`


dbmalkovsky (approved):
LGTM

PR #1952, opened 2024-02-01
[Draft] OP-1141 Implement credetials suggestion

emecas:
Here is an approach to  https://openhospital.atlassian.net/browse/OP-1141 using resources and a button and its ToolTipText.

I tried 3 different approach leaving the one in : `rsc/info_button3.html` that is the first part of the Tool-Tip-Text, then  it requires a `rsc/default_credential.properties` to verify it exists and taking the credentials text to be added as extra lines for the Tool-Tip-Text.

image

**TO DO:**

- [ ]  At the end of Login.java I have added 3new methods to manage resources that I need to relocate (where?) and optimize
       - public static String readFileToStringLineByLine(String path)
       - public static String readFileToString(String path) 
       - public static boolean verifyIfResourceExists(String path) 
- [ ] Also, a main method just for provisional testing that I will remove later

### DIFF
--- a/rsc/info_button.html
+++ b/rsc/info_button.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Default credentials</title>
+</head>
+<body>
+<h3>Default credentials</h3>
+    <h5>admin:admin</h5>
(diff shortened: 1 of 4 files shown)


emecas:
Closing this PR, replaced  by https://github.com/informatici/openhospital-gui/pull/1954 following suggestions

PR #1953, opened 2024-02-02
OP-1169 | DELETE last movement (only) in Pharmaceutical Stock Ward

ArnaudFonzam:
See OP-1169.


src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
```diff
@@ -410,6 +384,50 @@ private JButton getJButtonClose() {
...
+				}
+				int selectedRow = jTableOutcomes.getSelectedRow();
+				if (selectedRow == -1) {
+					MessageDialog.error(this, "angal.medicalstock.pleaseselectoucomemovement.msg");
+					return;
+				}
+				MovementWard selectedMovement = (MovementWard)jTableOutcomes.getValueAt(selectedRow, -1);
+				try {
+					MovementWard movWard = movWardBrowserManager.getLastMovementWard(wardSelected);
+					if (movWard.getDate().compareTo(selectedMovement.getDate()) == 0) {
```
mwithi, line 406: why comparing the date and not the id?

ArnaudFonzam: Because the Id is not présent in the outcome Table

mwithi: what do you mean? the table just shows a subset of information, the id should be in the `selectedMovement` object...

ArnaudFonzam: No. okay I have update the code and use the id like you suggest

bundle/language_en.properties
```diff
@@ -880,6 +880,7 @@ angal.medicalstock.pharmaceutical.col
 angal.medicalstock.pleasechooseatype.msg                                                               = Please choose a type.
 angal.medicalstock.pleaseselectamovement.msg                                                           = Please select a movement.
 angal.medicalstock.pleaseselectonlyonemovement.msg                                                     = Please select only one movement.
+angal.medicalstock.pleaseselectoucomemovement.msg                                                      = Please select a outcome movement.
```
dbmalkovsky, line 883:
```suggestion
angal.medicalstock.pleaseselectoucomemovement.msg                                                      = Please select an outcome movement.
```

dbmalkovsky (approved):
LGTM with one small grammar change.

src/main/java/org/isf/medicalstockward/gui/WardPharmacy.java
```diff
@@ -410,6 +384,50 @@ private JButton getJButtonClose() {
...
+					MessageDialog.error(this, "angal.medicalstock.pleaseselectoucomemovement.msg");
+					return;
+				}
+				MovementWard selectedMovement = (MovementWard)jTableOutcomes.getValueAt(selectedRow, -1);
+				try {
+					MovementWard movWard = movWardBrowserManager.getLastMovementWard(wardSelected);
+					if (movWard.getCode() == selectedMovement.getCode()) {
+						int delete = MessageDialog.yesNo(null, "angal.medicalstock.doyoureallywanttodeletethismovement.msg");
+						if (delete == JOptionPane.YES_OPTION) {
+							movWardBrowserManager.deleteLastMovementWard(movWard);
```
dbmalkovsky, line 409: Does this need to be in try-catch block in case an exception is thrown allowing for outputting a failure message?

mwithi: it is

mwithi: or do you mean another thing?

dbmalkovsky: I missed the outer try-catch; sorry.   I typically try to put them as narrow as possible but this is fine.   You can mark it resolved.

PR #1959, opened 2024-02-07
OP-1129 OP-1142 Script language update

mizzioisf:
See OP-1129 and OP-1142.

Startup script updates and fixes.

### DIFF
--- a/oh.sh
+++ b/oh.sh
@@ -77,7 +77,7 @@ DATABASE_PASSWORD="isf123"
 
 #######################  OH configuration  #########################
 # path and directories
-OH_DIR="oh"
+OH_DIR="."
 OH_DOC_DIR="doc"
 CONF_DIR="data/conf"
 DATA_DIR="data/db"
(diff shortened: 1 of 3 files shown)


oh.sh
```diff
@@ -1182,8 +1192,15 @@ function parse_user_input {
...
 			option="Z";
 		else
-			read -n 2 -p "Please select language [$OH_LANGUAGE_LIST]: " OH_LANGUAGE
+			echo ""
+			echo "Available languages:"
+			echo ""
+			# show all available languages
+			for ((i=0; i < ${#OH_LANGUAGE_LIST[@]}; i++ )); do echo " ${OH_LANGUAGE_LIST[$i]} - ${OH_LANGUAGE_LIST_INFO[$i]} "; done
+			echo ""
+			read -n 2 -p "Please select language [${OH_LANGUAGE_LIST[*]}]: " OH_LANGUAGE
```
dbmalkovsky, line 1201: There isn't a note about `en` being the default like there is in `oh.ps1` around line 1341.  Is that OK?

mwithi: Yes, also the message is different, just "Select language ... " (ref. line [1341](https://github.com/informatici/openhospital-gui/pull/1959/files#diff-f6027c8dacfb5e38348d01d05d1ae4625ec0c4d0ae3e96b3b943ddbce8a01a3cR1341))

mizzioisf: Fixed

mwithi (approved):
LGTM. Thanks a lot!

PR #1969, opened 2024-02-28
OP-1182 | Add Window to Manage global Inventory

ArnaudFonzam:
See OP-1182.
Paired with https://github.com/informatici/openhospital-core/pull/1280.

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -427,7 +427,7 @@ angal.common.qty.txt                                                            
 angal.common.quantity.txt                                                                              = Quantity
 angal.common.reminder.txt                                                                              = Reminder
 angal.common.result.txt                                                                                = Result
-angal.common.reference                                                                                 = Reference
+angal.common.reference.label                                                                           = Reference
 angal.common.save                                                                                      = Save
 angal.common.save.btn                                                                                  = Save
 angal.common.save.btn.key                                                                              = S
@@ -440,11 +440,6 @@ angal.common.selectsex.txt                                                      
 angal.common.sex.label                                                                                 = Sex:
 angal.common.sex.txt                                                                                   = Sex
 angal.common.sms.txt                                                                                   = SMS
-angal.common.status.txt                                                                                = Status
-angal.common.statecanceled                                                                             = Canceled                                 
-angal.common.stateinprogress                                                                           = In progress
-angal.common.statevalidate                                                                             = Validate
-angal.common.status.txt                                                                                = Status
 angal.common.stockcard.btn                                                                             = StockCard
 angal.common.stockcard.btn.key                                                                         = K
 angal.common.stockledger.btn                                                                           = Stock Ledger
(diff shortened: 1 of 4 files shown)


bundle/language_en.properties
```diff
@@ -427,6 +427,7 @@ angal.common.qty.txt
 angal.common.quantity.txt                                                                              = Quantity
 angal.common.reminder.txt                                                                              = Reminder
 angal.common.result.txt                                                                                = Result
+angal.common.reference                                                                                 = Reference
```
mwithi, line 430: this should be `.label` right?

src/main/java/org/isf/utils/jobjects/InventoryState.java
```diff
@@ -0,0 +1,55 @@
...
+package org.isf.utils.jobjects;
+
+
+public class InventoryState {
+	public InventoryState(){}
+	public enum State
+	 {		
+		PROGRESS ("1", "angal.common.statecanceled"),
+		CANCELED ("2", "angal.common.stateinprogress"),
+		VALIDATE ("3", "angal.common.statevalidate");
```
mwithi, line 31: the mapping looks wrong here...
Anyway I think that PROGRESS could be DRAFT
Then status for VALIDATE should be VALIDATED (past participle)

ArnaudFonzam: so I should update  stateinprogress to statedraft  and angal.common.statevalidate to angal.common.statevalidated. right?

src/main/java/org/isf/medicalinventory/gui/InventoryBrowser.java
```diff
@@ -0,0 +1,449 @@
...
+		panelFooter = getPanelFooter();
+		getContentPane().add(panelFooter, BorderLayout.SOUTH);
+
+		ajustWidth();
+
+		pagesCombo.setEditable(true);
+		previous.setEnabled(false);
+		next.setEnabled(false);
+		next.addActionListener(actionEvent -> {
+				if (!previous.isEnabled())
```
dbmalkovsky, line 115: missing braces

src/main/java/org/isf/medicalinventory/gui/InventoryBrowser.java
```diff
@@ -0,0 +1,449 @@
...
+				START_INDEX += PAGE_SIZE;
+				jTableInventory.setModel(new InventoryBrowsingModel(START_INDEX, PAGE_SIZE));
+				if ((START_INDEX + PAGE_SIZE) > TOTAL_ROWS) {
+					next.setEnabled(false);
+				}
+				pagesCombo.setSelectedItem(START_INDEX / PAGE_SIZE + 1);
+		});
+
+		previous.addActionListener(actionEvent -> {
+				if (!next.isEnabled())
```
dbmalkovsky, line 126: missing braces

bundle/language_en.properties
```diff
@@ -440,6 +441,10 @@ angal.common.sex.label
 angal.common.sex.txt                                                                                   = Sex
 angal.common.sms.txt                                                                                   = SMS
 angal.common.status.txt                                                                                = Status
+angal.common.statecanceled                                                                             = Canceled                                 
+angal.common.stateinprogress                                                                           = In progress
+angal.common.statevalidate                                                                             = Validate
+angal.common.status.txt                                                                                = Status
```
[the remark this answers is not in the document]
dbmalkovsky: And why don't they have have the proper suffixes on the keys?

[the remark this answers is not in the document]
ArnaudFonzam: so I should use angal.inventory.statecanceled, angal.inventory.statedraft and angal.inventory.statevalidated

src/main/java/org/isf/utils/jobjects/InventoryState.java
mwithi: for LaboratoryStatus we did just like this. 
If the lowercase part is the one that will go in the database, we SHOULD NOT use translations here.
```
package org.isf.lab.model;

public enum LaboratoryStatus {
	draft, open, done, invalid, deleted;
}
```
The statuses you planned for inventory are instead: `draft`, `validated` and `canceled`, correct? ("draft" is to keep consistency with other models in CORE).
Then, translations are related to GUI, and I guess you should use `angal.medicalstock.inventory` key or similar.

mwithi (approved):
LGTM. Thanks a lot!

PR #1973, opened 2024-03-19
OP-1191 | Create window to Add new ward Inventory

JantBogard:
See OP-1191.


src/main/java/org/isf/medicalinventory/gui/InventoryWardEdit.java
```diff
@@ -0,0 +1,505 @@
...
+            return pColums.length;
+        }
+
+        @Override
+        public boolean isCellEditable(int rowIndex, int columnIndex) {
+            return columnEditable[columnIndex];
+        }
+    }
+
+    public String formatDateTime(GregorianCalendar time) {
```
dbmalkovsky, line 366: Where is this method used?   And I believe 'GregorianCalendar` is no longer used in the project.

PR #1979, opened 2024-04-06
OP-1149 Update to Spring Boot 3.2.x

dbmalkovsky:
See OP-1149

Note the **main** class is now `org.isf.Application` and not `org.isf.menu.gui.Menu`.   The reason for the change is because Spring Boot scans downward from the main class looking for beans, etc.

`ApplicationContext.xml` is no longer used (Spring has been moving away from xml files).  The information is now included in either `application.properties` or `application.yml`. Much of the information is gleaned from the class files themselves by scanning downward from the main class.

The CORE PR companion is:  https://github.com/informatici/openhospital-core/pull/1303


mwithi (approved):
LGTM. Thanks a lot!

PR #1985, opened 2024-04-17
OP-1149 The main class is no longer `org.isf.menu.gui.Menu` but `org.isf.Application`

dbmalkovsky:
See OP-1149

The main class is no longer `org.isf.menu.gui.Menu` but `org.isf.Application` so the Spring Boot class scanning will work.  The main class needs to be at the top of the tree for best results.


mizzioisf:
Great ! Well done @dbmalkovsky, thank you !

mizzioisf (approved):
LGTM ! Thank you !

PR #1987, opened 2024-04-22
OP-1210 quick "fix" for changes from CORE PR-1210.

dbmalkovsky:
See OP-1210

Without this temporary change the test will not compile.

Will make the **correct** fix as soon as I have the time; sorry for the breakage.


mwithi (approved):
LGTM. Thanks a lot!

PR #1993, opened 2024-05-09
OP-1229 Medicals sorted by name in Pharmaceutical Stock

mwithi:
See OP-1229.


dbmalkovsky (approved):
LGTM even if the CI builds fail.

mwithi:
> LGTM even if the CI builds fail.

yeah, that is because the gui is branched from 1.14.1 but CI is always against actual develop 👍
I need to merge these changes also into develop once 1.14.2 is released

PR #2001, opened 2024-05-10
OP-1222 ProductLedger improvements

mwithi:
See OP-1222.


dbmalkovsky (approved):
LGTM given my very limited knowledge about jrxml

PR #2028, opened 2024-07-25
OP-1186 View Inventory

ArnaudFonzam:
See OP-1186.


src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -215,11 +214,15 @@ private void fireInventoryUpdated() {
 	private MedicalDsrStockMovementTypeBrowserManager movTypeManager = Context.getApplicationContext().getBean(MedicalDsrStockMovementTypeBrowserManager.class);
 	private SupplierBrowserManager supplierManager = Context.getApplicationContext().getBean(SupplierBrowserManager.class);
 	private WardBrowserManager wardManager = Context.getApplicationContext().getBean(WardBrowserManager.class);
-
 	
 	public InventoryEdit() {
 		mode = "new";
 		initComponents();
+		if (mode.equals("view")) {
```
mwithi: How can `mode` be different from `"new"` at this point?

mwithi: the real `"view"` mode is passed in the other constructor at `line 232` and the checks to enable or disable buttons are already done in the `initComponents()` method, so this check here is useless and the OP-1186 could be already working in https://github.com/informatici/openhospital-gui/tree/OP-1199-Main-Store-Inventory itself

mwithi (approved):
LGTM. Thanks a lot!

PR #2041, opened 2024-08-20
OP-1302 | Make JTable sortable in edit inventory

ArnaudFonzam:
See OP-1302.


mwithi:
Looks like that when sorting by UNIT PRICE it shows `Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.math.BigDecimal (java.lang.Integer and java.math.BigDecimal are in module java.base of loader 'bootstrap')`

when sorting by TOTAL PRICE it shows `Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Double (java.lang.Integer and java.lang.Double are in module java.base of loader 'bootstrap')`

mwithi (approved):
LGTM. Thanks a lot!

PR #2042, opened 2024-08-21
OP-1187 | Printing Inventory

loique70:
See OP-1187


src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -865,8 +874,44 @@ private JButton getCloseButton() {
...
-	
-	private JButton getCleanTableButton() {
+
+	private JButton getPrintButton() {
+		printButton = new JButton(MessageBundle.getMessage("angal.common.print.btn"));
+		printButton.setMnemonic(MessageBundle.getMnemonic("angal.common.print.btn.key"));
+		printButton.setEnabled(true);
+
+		printButton.addActionListener(e -> {
+			if (inventory == null || inventory.getId() <= 0) {
```
mwithi, line 884: This is more type safe because getId() might be null and unboxing could return a NPE

```suggestion
			if (inventory == null || inventory.getId() == null || inventory.getId() <= 0) {
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -865,8 +874,44 @@ private JButton getCloseButton() {
...
+				MessageDialog.info(this, "angal.inventory.pleasesavebeforprinting");
+				return;
+			}
+
+			if (InventoryStatus.done.name().equals(inventory.getStatus())) {
```
mwithi, line 889: Why to print only when DONE status? This make non sense, maybe you meant the opposite?
In any case why to depend on the status? Please elaborate a bit.

```suggestion
			if (!InventoryStatus.done.name().equals(inventory.getStatus())) {
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -865,8 +874,44 @@ private JButton getCloseButton() {
...
+					printQtyReal = 1;
+				}
+
+				try {
+					JasperReportResultDto reportResult = jasperReportsManager.getInventoryReportPdf(inventory, "Inventory", printQtyReal);
+					if (reportResult != null) {
+						MessageDialog.info(this, "angal.inventory.print.success.msg");
+					} else {
+						MessageDialog.error(this, "angal.inventory.printing.error.msg");
+					}
```
mwithi, line 902: I don't get this: the report should be shown to the user before printing (not only created as PDF) and in case the report is OK the user can decide to print or not from the interface (see `INTERNALVIEWER` and other reports in general).

There's no need of a `print.success.msg`.

mwithi (changes requested):
In addition to other suggestions, the resulting report has overlapped fields:

image

mwithi:
Closing for new PR on the way

PR #2049, opened 2024-09-06
OP-1206 Update script to use the new application.properties.dist file

mwithi:
See OP-1206.
Paired with https://github.com/informatici/openhospital-api/pull/465.


mizzioisf (approved):
Good, go ahead and merge.

PR #2060, opened 2024-10-02
OP-1344: dialog method takes bundle key and not the final text

dbmalkovsky:
Found while testing new jars


mwithi (approved):
LGTM. Thanks a lot!

PR #2061, opened 2024-10-03
OP-1352 Fix one label

mwithi:
See OP-1352.
Paired with https://github.com/informatici/openhospital-core/pull/1428.


dbmalkovsky (approved):
LGTM

PR #2071, opened 2024-10-23
worfklow: trying to build against a core branch with same name (develop otherwise)

mwithi:
Same as https://github.com/informatici/openhospital-api/pull/484.


dbmalkovsky (approved):
LGTM

PR #2074, opened 2024-10-28
OP-1339 Fix lots overlooking when they reach zero quantity

mwithi:
See OP-1339.
Paired with https://github.com/informatici/openhospital-core/pull/1441.

Also add workflow refactoring

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -829,10 +829,13 @@ angal.medicalstock.multiplecharging.chargetype                                  
 angal.medicalstock.multiplecharging.chooseamedical                                                     = Choose a Medical
 angal.medicalstock.multiplecharging.cost                                                               = Cost
 angal.medicalstock.multiplecharging.datecannotbebeforelastmovementdate.msg                             = Date cannot be before last movement date.
+angal.medicalstock.multiplecharging.empty                                                              = Empty
 angal.medicalstock.multiplecharging.existinglot                                                        = Existing Lots
+angal.medicalstock.multiplecharging.expired                                                            = Expired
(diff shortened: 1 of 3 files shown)


mwithi:
going to fix build upon "push" events too...

mwithi:
@dbmalkovsky now should be ok

bundle/language_en.properties
```diff
@@ -829,10 +829,13 @@ angal.medicalstock.multiplecharging.chargetype
 angal.medicalstock.multiplecharging.chooseamedical                                                     = Choose a Medical
 angal.medicalstock.multiplecharging.cost                                                               = Cost
 angal.medicalstock.multiplecharging.datecannotbebeforelastmovementdate.msg                             = Date cannot be before last movement date.
+angal.medicalstock.multiplecharging.empty                                                              = Empty
```
dbmalkovsky, line 832:
```suggestion
angal.medicalstock.multiplecharging.empty.txt                                                          = Empty
```
Technically all message constants should have some kind of suffix: txt, btn, key, msg, fmt.msg, etc

See `expired` and `hideemptylots` below for other examples.

dbmalkovsky (approved):
LGTM (given you ignore my suffix suggestion).

bundle/language_en.properties
mwithi: I realize now the all the other bundles around have no suffix... but in this case what suffix we should apply?

ArnaudFofou (approved):

bundle/language_en.properties
mwithi: not sure about the suffix to use, we will do a revision in another issue.

PR #2075, opened 2024-10-29
OP-1340 Fix when medical has no previous lots

mwithi:
See OP-1340.
The point was the missing Medical on a new lot obtained with `askLot` method. We miss the cost also.
So I tried to rearrange the code in a more (hopefully) functional way.

### DIFF
--- a/src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
+++ b/src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
@@ -467,7 +467,7 @@ public class MovStockMultipleCharging extends JDialog {
 		newLot.setMedical(med);
 
 		if (!setOrValidateCost(newLot, qty)) {
-			return null; // Exit if cost was invalid
+			return null;
 		}
 		return newLot;
 	}
--- a/src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
+++ b/src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
@@ -436,13 +436,16 @@ public class MovStockMultipleCharging extends JDialog {
 								if (lot == null) {
 									return;
 								}
+								// Lot Cost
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
```diff
@@ -470,6 +460,34 @@ private JTextField getJTextFieldSearch() {
...
 	}
 
+	private Lot createNewLot(Medical med, int qty) {
+		LocalDateTime preparationDate = TimeTools.getNow().truncatedTo(ChronoUnit.MINUTES);
+		LocalDateTime expiringDate = askExpiringDate();
+		Lot newLot = new Lot("", preparationDate, expiringDate);
+		newLot.setMedical(med);
+
+		if (!setOrValidateCost(newLot, qty)) {
+			return null; // Exit if cost was invalid
```
mwithi:
```suggestion
			return null;
```

src/main/java/org/isf/medicalstock/gui/MovStockMultipleCharging.java
```diff
@@ -470,6 +460,34 @@ private JTextField getJTextFieldSearch() {
...
+		}
+		return newLot;
+	}
+
+	private boolean setOrValidateCost(Lot lot, int qty) {
+		BigDecimal cost = lot.getCost() != null ? lot.getCost() : BigDecimal.ZERO;
+		if (GeneralData.LOTWITHCOST && (cost.equals(BigDecimal.ZERO) || lot.getCost() == null)) {
+			cost = askCost(qty);
+			if (cost.compareTo(BigDecimal.ZERO) == 0) {
+				return false; // Invalid cost, exit the method
```
mwithi:
```suggestion
				return false;
```

ArnaudFofou (approved):

dbmalkovsky (approved):
LGTM

PR #2088, opened 2024-11-19
OP-1361 | Fix getMedicalsWard method

ArnaudFofou:
See OP-1361.
Paired with https://github.com/informatici/openhospital-core/pull/1452.


mwithi (approved):
LGTM. Thanks a lot!

PR #2091, opened 2024-11-21
OP-1304 | Enhance product selections in edit inventory

ArnaudFofou:
See OP-1304.


src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -486,6 +503,52 @@ private GoodDateChooser getJCalendarFrom() {
 		return jCalendarInventory;
 	}
 
+	private JButton getSelectedButton() {
+		if (selectButton == null) {
+			selectButton = new JButton(MessageBundle.getMessage("angal.common.select.btn"));
+			selectButton.setMnemonic(MessageBundle.getMnemonic("angal.common.select.btn.key"));
+			selectButton.addActionListener(actionEvent -> {
+				specificRadio.setSelected(false);
+		        mainPanel = new JPanel();
```
dbmalkovsky: the formatting is off starting here

ArnaudFofou: I don't have the same formatting localy
bug

dbmalkovsky: To quote @mwithi it doesn't matter how it looks locally but how it looks on github:
image

You can see this by viewing the file in the changed files list on the right side:
image

FWIW, the link is: https://github.com/informatici/openhospital-gui/blob/9fb1f94bbc1c6e47025192d9950fa0bbbd9b9cb2/src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+    		radioButtonAll.setSelected(true);
+    	}
+    	radioButtonAll.addActionListener(actionEvent -> {
+    		if (radioButtonAll.isSelected()) {
+    			radioOnlyNonZero.setSelected(false);
+				radioWithMovement.setSelected(false);
+			}
+		});
+		return radioButtonAll;
+	}
```
mwithi, line 2185: Fix alignment

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+    		radioOnlyNonZero.setMnemonic(MessageBundle.getMnemonic("angal.inventory.medicalwithonlynonzeroqty.btn.key"));
+    	}
+    	radioOnlyNonZero.addActionListener(actionEvent -> {
+			if (radioOnlyNonZero.isSelected()) {
+				radioButtonAll.setSelected(false);
+				radioWithMovement.setSelected(false);
+			}
+		});
+		return radioOnlyNonZero;
+	}
```
mwithi, line 2199: Fix alignment

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+			}
+			
+			medicalTypeComboBox.addActionListener(actionEvent -> {
+				medicalTypeSelected = (MedicalType) medicalTypeComboBox.getSelectedItem();
+			});
+		}
+		return medicalTypeComboBox;
+	}
+	
+    private JRadioButton getAllRadioButton() {
```
mwithi: Fix indentation for all these buttons:
`getAllRadioButton`
`getMedicalWithNonZeroQuatityRadioButton`
`getMedicalWithMovementRadioButton`
`getCancelButton`

and leave an emtpy line after each `}`

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+    
+	private List<MedicalInventoryRow> loadNewInventoryTable(MedicalType medicalTypeSelected, boolean withMovement) throws OHServiceException {	
+		List<MedicalInventoryRow> inventoryRowsList = getMedicalInventoryRowsWithMovement();
+		if (medicalTypeSelected != null) {
+			inventoryRowsList = inventoryRowsList.stream().filter(inv -> inv.getMedical().getType().getDescription().equals(medicalTypeSelected.getDescription())).collect(Collectors.toList());
+		}
+		return inventoryRowsList;
+	}
+    
+    private JButton getOkButton() {
```
mwithi: also this `getOkButton` doesn't look indented correctly

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -71,6 +73,8 @@
 import javax.swing.event.EventListenerList;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
+import javax.swing.event.TableModelEvent;
+import javax.swing.event.TableModelListener;
```
mwithi: these two import are not used

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+            							specificRadio.setSelected(true);
+            						}
+
+            					} else {
+            						try {
+        								jTableInventoryRow.setModel(new InventoryRowModel(true));
+        							} catch (OHServiceException e) {
+        								OHServiceExceptionUtil.showMessages(e);
+        							}
+            					}
```
mwithi: `jTableInventoryRow.setModel(new InventoryRowModel(true));
`
is executed in both branches of the `if-else` block, maybe you can simplify?

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+    private JButton getOkButton() {
+    	if (jButtonOk == null) {
+    		jButtonOk = new JButton(MessageBundle.getMessage("angal.common.ok.btn"));
+    		jButtonOk.setMnemonic(MessageBundle.getMnemonic("angal.common.ok.btn.key"));
+    		jButtonOk.addActionListener(actionEvent -> {
+        		try {
+        			medicalTypeSelected = (MedicalType) medicalTypeComboBox.getSelectedItem();
+            		if (radioButtonAll.isSelected()) {
+            			if (medicalTypeSelected.getDescription().equals(MessageBundle.getMessage("angal.common.all.txt"))) {
+            				if (!areAllMedicalsInInventory()) {
```
mwithi: The method `areAllMedicalsInInventory` is called 10 times in this class and each time retrieves all medicals from the DB... maybe this could be improved and reduce the traffic with the DB?

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+            					fireInventoryUpdated();
+            					code = null;
+            					adjustWidth();
+            					jButtonCancel.doClick();
+            					MessageDialog.info(null, "angal.invetory.tablehasbeenupdated.msg");
+            				} else {
+            					jButtonCancel.doClick();
+        						MessageDialog.info(null, "angal.inventory.youhavealreadyaddedallproduct.msg");
+        					}
+            			}
```
mwithi: Same here:` jTableInventoryRow.setModel(new InventoryRowModel(medicalTypeSelected));` used in both cases of the `if-else` statement

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+            					code = null;
+            					adjustWidth();
+            					jButtonCancel.doClick();
+            					if (inventoryRowList.size() > 0) {
+            						MessageDialog.info(null, "angal.invetory.tablehasbeenupdated.msg");	
+            					}
+            				} else {
+            					jButtonCancel.doClick();
+        						MessageDialog.info(null, "angal.inventory.youhavealreadyaddedallproduct.msg");
+        					}
```
mwithi: Also here for `jTableInventoryRow.setModel(new InventoryRowModel(true, null));`

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+            			} else {
+            				if (!areAllMedicalsInInventory()) {
+            					codeTextField.setEnabled(false);
+            					codeTextField.setText("");
+            					if (inventoryRowSearchList.size() > 0) {
+            						int info = MessageDialog.yesNo(null, "angal.inventoryrow.doyouwanttoaddallnotyetlistedproducts.msg");
+            						if (info == JOptionPane.YES_OPTION) {
+            							jTableInventoryRow.setModel(new InventoryRowModel(true, medicalTypeSelected));
```
mwithi: Here too with `jTableInventoryRow.setModel(new InventoryRowModel(true, medicalTypeSelected));`

mwithi (changes requested):
Although the feature is doing what expected, there is room of code improvements, please read the comments

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: the difference between the block if and the block else if the line 2257 (ask user if he need to add all not yet listed products)

mwithi: so, following the DRY principle, the message is the only thing should be under if-else block...

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -486,6 +500,52 @@ private GoodDateChooser getJCalendarFrom() {
...
+		        mainPanel.add(leftPanel, BorderLayout.WEST);
+		        mainPanel.add(rightPanel, BorderLayout.EAST);
+		        mainPanel.add(bottomPanel, BorderLayout.SOUTH);
+		        
+		        frame = new JFrame();
+		        frame.add(mainPanel);
+		        frame.setSize(450, 200);
+		        frame.setTitle(MessageBundle.getMessage("angal.inventoryrow.lotinformation.title"));
+		        frame.setLocationRelativeTo(null);
+		        frame.setDefaultCloseOperation(JFrame.EXIT_ON_CLOSE);
```
dbmalkovsky:
```suggestion
		        frame.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
```
Use the base class.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+		if (withNonZeroQty) {
+			inventoryRowsList = inventoryRowsList.stream().filter(inv -> inv.getTheoreticQty() > 0).collect(Collectors.toList());
+		}
+		if (medicalTypeSelected != null) {
+			inventoryRowsList = inventoryRowsList.stream().filter(inv -> inv.getMedical().getType().getDescription().equals(medicalTypeSelected.getDescription())).collect(Collectors.toList());
+		}
+		return inventoryRowsList;
+	}
+    
+	private List<MedicalInventoryRow> loadNewInventoryTable(MedicalType medicalTypeSelected, boolean withMovement) throws OHServiceException {
```
dbmalkovsky: I don't think the method parameter, `withMovement` is used in this method and thus can be removed

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
 		return false;
 	}
 	
+	private JComboBox<MedicalType> getJComboMedicalType() {
+		if (medicalTypeComboBox == null) {
+			medicalTypeComboBox = new JComboBox<MedicalType>();
```
dbmalkovsky:
```suggestion
			medicalTypeComboBox = new JComboBox<>();
```
Don't use the type specification in the constructor call.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2025,6 +2148,306 @@ private boolean checkParameters(String wardCode, String chargeCode, String disch
...
+    		jButtonOk.setMnemonic(MessageBundle.getMnemonic("angal.common.ok.btn.key"));
+    		jButtonOk.addActionListener(actionEvent -> {
+        		try {
+        			medicalTypeSelected = (MedicalType) medicalTypeComboBox.getSelectedItem();
+            		if (radioButtonAll.isSelected()) {
+            			if (medicalTypeSelected.getDescription().equals(MessageBundle.getMessage("angal.common.all.txt"))) {
+            				if (!areAllMedicalsInInventory()) {
+            					codeTextField.setEnabled(false);
+            					codeTextField.setText("");
+            					if (inventoryRowSearchList.size() > 0) {
```
dbmalkovsky: `isEmpty()` should be used to test for emptiness of a collection.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1668,6 +1714,76 @@ private List<MedicalInventoryRow> getMedicalInventoryRows(String code) throws OH
...
+						inventoryRowsList.add(inventoryRowTemp);
+					}
+				}
+			}
+		}
+		return inventoryRowsList;
+	}
+	
 	private void addInventoryRow(String code) throws OHServiceException {
 		List<MedicalInventoryRow> inventoryRowsList = new ArrayList<MedicalInventoryRow>();
```
dbmalkovsky:
```suggestion
		List<MedicalInventoryRow> inventoryRowsList = new ArrayList<>();
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1668,6 +1714,76 @@ private List<MedicalInventoryRow> getMedicalInventoryRows(String code) throws OH
...
 		List<Medical> medicalList = new ArrayList<Medical>();
```
dbmalkovsky:
```suggestion
		List<Medical> medicalList = new ArrayList<>();
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: okay

mwithi (approved):

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2010,23 +2137,290 @@ private void resetVariable() {
...
+			radioButtonAll = new JRadioButton(MessageBundle.getMessage("angal.common.all.btn"));
+			radioButtonAll.setMnemonic(MessageBundle.getMnemonic("angal.common.all.btn.key"));
+			radioButtonAll.setSelected(true);
+		}
+		radioButtonAll.addActionListener(actionEvent -> {
+			if (radioButtonAll.isSelected()) {
+				radioOnlyNonZero.setSelected(false);
+				radioWithMovement.setSelected(false);
+			}
+		});
```
mwithi: In java swing this logic can be avoided by adding the three `JRadioButtons` to the same `ButtonGroup`

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2010,23 +2137,290 @@ private void resetVariable() {
...
+							.collect(Collectors.toList());
+		}
+		return inventoryRowsList;
+	}
+
+	private JButton getOkButton() {
+		if (jButtonOk == null) {
+			jButtonOk = new JButton(MessageBundle.getMessage("angal.common.ok.btn"));
+			jButtonOk.setMnemonic(MessageBundle.getMnemonic("angal.common.ok.btn.key"));
+			jButtonOk.addActionListener(actionEvent -> {
```
mwithi, line 2171: don't you realize that the three `if` blocks are identical now? The three options have same effect...

```
if (radioButtonAll.isSelected()) {
    // same logic with new InventoryRowModel(true);
}
if (radioOnlyNonZero.isSelected()) {
    // same logic with new InventoryRowModel(true)
}
if (radioWithMovement.isSelected()) {
    // same logic with new InventoryRowModel(true)
}
```

Moreover, I don't think we still need to manage the `specificRadio` button because now it's the only JRadioButton since the other `selectButton` is now a standard JButton; the `codeTextField` (always active) should be enough with its `KeyListener` to allow the user to add at any time specific products to the inventory, so the JTextField + a JLabel is enough, see issue's screenshot as example.

mwithi (changes requested):
The new logic needs to be fixed

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: oops sory I used now the right constructor for differents cases

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: okay I have remove the that

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2010,23 +2137,245 @@ private void resetVariable() {
...
+	private JRadioButton getAllRadioButton() {
+		if (radioButtonAll == null) {
+			radioButtonAll = new JRadioButton(MessageBundle.getMessage("angal.common.all.btn"));
+			radioButtonAll.setMnemonic(MessageBundle.getMnemonic("angal.common.all.btn.key"));
+			radioButtonAll.setSelected(true);
+		}
+		return radioButtonAll;
+	}
+
+	private JRadioButton getMedicalWithNonZeroQuatityRadioButton() {
```
mwithi:
```suggestion
	private JRadioButton getMedicalWithNonZeroQuantityRadioButton() {
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2010,23 +2137,245 @@ private void resetVariable() {
...
+	private JRadioButton getAllRadioButton() {
+		if (radioButtonAll == null) {
+			radioButtonAll = new JRadioButton(MessageBundle.getMessage("angal.common.all.btn"));
+			radioButtonAll.setMnemonic(MessageBundle.getMnemonic("angal.common.all.btn.key"));
+			radioButtonAll.setSelected(true);
+		}
+		return radioButtonAll;
+	}
+
+	private JRadioButton getMedicalWithNonZeroQuantityRadioButton() {
```
dbmalkovsky, line 2121: I don't see where this `private` method is ever used.....

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1186,13 +1238,66 @@ public InventoryRowModel() throws OHServiceException {
...
+			if (areAllMedicalsInInventory()) {
+				specificRadio.setSelected(false);
+				codeTextField.setEnabled(false);
+			} else {
+				specificRadio.setSelected(true);
+				codeTextField.setEnabled(true);
+			}
+		}
+
+		public InventoryRowModel(MedicalType medType, boolean withMovement) throws OHServiceException {
```
dbmalkovsky, line 1237: the method parameter, `withMovement` is never used....

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1668,9 +1749,49 @@ private List<MedicalInventoryRow> getMedicalInventoryRows(String code) throws OH
...
+		List<Medical> medicalListWithMovement = new ArrayList<>();
+		List<Lot> lots = null;
+		MedicalInventoryRow inventoryRowTemp = null;
+		List<Medical> medicalList = medicals;
+		ListIterator<Medical> medicalListIterator = medicalList.listIterator();
+		while (medicalListIterator.hasNext()) {
+			Medical med = medicalListIterator.next();
+			Integer medicalCodde = med.getCode();
+			List<Movement> movements = movBrowserManager.getMovements(medicalCodde, null, null, null, null, null, null, null, null, null);
+			if (movements.size() > 0) {
```
dbmalkovsky: use `isEmpty()` instead; true for lines 1072 and 1734 also

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1668,9 +1749,49 @@ private List<MedicalInventoryRow> getMedicalInventoryRows(String code) throws OH
...
+				medicalListWithMovement.add(med);
+			}
+		}
+		medicalListIterator = medicalListWithMovement.listIterator();
+		while (medicalListIterator.hasNext()) {
+			Medical med = medicalListIterator.next();
+			lots = movStockInsertingManager.getLotByMedical(med, false);
+			double actualQty = med.getInqty() - med.getOutqty();
+			if (lots.size() == 0) {
```
dbmalkovsky: use `isEmpty()` here too

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -253,6 +271,11 @@ public InventoryEdit(MedicalInventory inventory, String mod) {
 	private void initComponents() {
 		inventoryRowList = new ArrayList<>();
 		inventoryRowSearchList = new ArrayList<>();
+		try {
+			medicals = medicalBrowsingManager.getMedicals();
+		} catch (OHServiceException e) {
+			OHServiceExceptionUtil.showMessages(e);
+		}
 		setDefaultCloseOperation(JFrame.DO_NOTHING_ON_CLOSE);
```
dbmalkovsky:
```suggestion
		setDefaultCloseOperation(WindowConstants.DO_NOTHING_ON_CLOSE);
```

dbmalkovsky:
There needs to be a review of the  whole file of `InventoryEdit` not just the recent changes.  For example, there are other occurrences that should be using `isEmtpy()`; there are allocations where `<>` should be used instead of the type; there are occurrences where Boxed types (specially `Integer`) is being compared with `==` instead of `equals()`, etc.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: It is use at line 2352

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
mwithi: My comment included:

"Moreover, I don't think we still need to manage the `specificRadio` button because now it's the only JRadioButton since the other `selectButton` is now a standard JButton; the `codeTextField` (always active) should be enough with its `KeyListener` to allow the user to add at any time specific products to the inventory, so the JTextField + a JLabel is enough, see issue's screenshot as example."

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1145,35 +1205,29 @@ class InventoryRowModel extends DefaultTableModel {
...
 			}
 			if (areAllMedicalsInInventory()) {
-				allRadio.setSelected(true);
 				specificRadio.setSelected(false);
 				codeTextField.setEnabled(false);
 			} else {
 				specificRadio.setSelected(true);
 				codeTextField.setEnabled(true);
-				allRadio.setSelected(false);
 			}
```
mwithi: By effect of my previous comment, this block can be removed in each constructor.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -1145,35 +1205,29 @@ class InventoryRowModel extends DefaultTableModel {
...
 		}
 
 		public InventoryRowModel() throws OHServiceException {
+			inventoryRowList.clear();
 			if (!inventoryRowSearchList.isEmpty()) {
 				inventoryRowSearchList.clear();
 			}
```
mwithi: what is the reason of this `if`? Just clear the `inventoryRowSearchList` if that is the aim, isnt't it?

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
mwithi: Moreover, this method has still a lot of duplicated code, are you sure it cannot be further refactored?

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -484,6 +508,57 @@ private GoodDateChooser getJCalendarFrom() {
...
+				
+				// Map actions to buttons
+				initialiseActions();
+				// Convertir la HashMap en TreeMap pour trier les clés
+				Map<AbstractButton, Runnable> sortedActionMap = new TreeMap<>(Comparator.comparing(AbstractButton::getText));
+				sortedActionMap.putAll(actions);
+				// Add ActionListener to each button
+				 sortedActionMap.forEach((key, value) ->  {
+		            rightPanel.add(key);
+		        });
```
mwithi:
```suggestion
				sortedActionMap.forEach((key, value) ->  {
		            rightPanel.add(key);
		        });
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -67,13 +72,15 @@
...
 import javax.swing.SwingUtilities;
+import javax.swing.WindowConstants;
 import javax.swing.border.EmptyBorder;
 import javax.swing.event.EventListenerList;
 import javax.swing.event.ListSelectionEvent;
 import javax.swing.event.ListSelectionListener;
 import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.DefaultTableModel;
 
+import org.hibernate.internal.build.AllowSysOut;
```
mwithi: This import is not used

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2036,4 +2196,125 @@ private  boolean areAllMedicalsInInventory() throws OHServiceException {
 		}
 		return false;
 	}
-}
\ No newline at end of file
+	
+	private void initialiseActions() {
```
mwithi: I understand this attempt of extract actions from the `okButton` logic. Nevertheless the code duplication is among the three radio button actions themselves: every action only differ by its core part `jTableInventoryRow.setModel(new InventoryRowModel(...args));`; and the message in case of succcess, all the remaining code is identical and repeated twice for each action, for a total of 6 duplicated blocks:

Can you see the same pattern I do?
`action -> updateModel, successMessage`

Anyway, let's fix the method name:
```suggestion
	private void initializeActions() {
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
mwithi, line 1: Please add an empty line before `addMedInRowInInventorySearchList` and `resetVariable` methods.

image

mwithi (approved):

PR #2092, opened 2024-11-21
OP-1341 Enable Medical Type for Soft Deletion

mohit10696:
See OP-1341.
Paired with https://github.com/informatici/openhospital-core/pull/1458.

### DIFF
--- a/src/main/java/org/isf/medtype/gui/MedicalTypeBrowserEdit.java
+++ b/src/main/java/org/isf/medtype/gui/MedicalTypeBrowserEdit.java
@@ -26,6 +26,7 @@ import java.awt.BorderLayout;
 import java.util.EventListener;
 
 import javax.swing.JButton;
+import javax.swing.JCheckBox;
 import javax.swing.JDialog;
 import javax.swing.JFrame;
 import javax.swing.JLabel;
@@ -95,8 +96,10 @@ public class MedicalTypeBrowserEdit extends JDialog {
 	private JButton cancelButton;
 	private JButton okButton;
 	private JTextField descriptionTextField;
+	private JCheckBox activeCheckbox;
 	private VoLimitedTextField codeTextField;
 	private String lastdescription;
+	private char isLastDeleted;
 	private MedicalType medicalType;
 	private boolean insert;
 	private JPanel jDataPanel;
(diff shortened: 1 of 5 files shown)


mohit10696:
image

image

src/main/java/org/isf/medtype/gui/MedicalTypeBrowserEdit.java
```diff
@@ -263,6 +270,14 @@ private JTextField getCodeTextField() {
 		}
 		return codeTextField;
 	}
+	
+	private JCheckBox getActiveField() {
+		if (activeCheckbox == null) {
+			activeCheckbox = new JCheckBox("Active");
```
dbmalkovsky: shouldn't this text be localized?

src/main/java/org/isf/medtype/gui/MedicalTypeBrowserEdit.java
```diff
@@ -276,7 +291,9 @@ private JPanel getJDataPanel() {
 			jDataPanel.add(getCodeTextField());
 			jDataPanel.add(new JLabel(MessageBundle.getMessage("angal.common.description.txt") + ':'));
 			jDataPanel.add(getDescriptionTextField());
-			SpringUtilities.makeCompactGrid(jDataPanel, 2, 2, 5, 5, 5, 5);
+			jDataPanel.add(new JLabel(MessageBundle.getMessage("Active") + ':'));
```
dbmalkovsky: "Active" isn't a typical message bundle string key; they start with `angal.`

src/main/java/org/isf/medtype/gui/MedicalTypeBrowserEdit.java
mwithi: I would not use "Active" but rather "Deleted" (so default is `false`) as in the rest of the application

mohit10696:
@dbmalkovsky , Can someone of you confirm about the text ? 
Do you want as active over there or deleted over there? After that, I will localize that text.

mwithi:
> Can someone of you confirm about the text ? Do you want as active over there or deleted over there? After that, I will localize that text.

Since we have already fields like XXX_ACTIVE in the DB (for future uses), I would use "Deleted" (like the new field) to avoid confusion.

mohit10696:
@mwithi  Refactored active to deleted text and corresponding logic. Hoping to get  PR merged 🤞

mwithi (approved):
LGTM. Thanks a lot!

PR #2094, opened 2024-11-22
OP-1302 | Make JTable sortable in edit inventory

ArnaudFofou:
See OP-1302.


src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
mwithi, line 1: Please consider model/view index convertions: see https://stackoverflow.com/questions/4151850/how-to-keep-track-of-row-index-when-jtable-has-been-sorted-by-the-user

ArnaudFofou: okay thank I will use that idea to customise  JTableRowSorter

PR #2099, opened 2024-12-03
Chore: MessageDialog methods takes a bundle key not the resolved message

dbmalkovsky:
Makes one wonder about testing ......


mwithi (approved):
LGTM. Thanks a lot!

mwithi:
> Makes one wonder about testing ......

It is strange, but the develop version works fine, do yo want to check before we merge?

image

dbmalkovsky:
It works because of the code can't find the key in the bundle it just returns the same string back as the result.   I bet if you look in the log you'll find an error message. 

Sure enough:   `04/Dec/2024 12:47:07] [admin:admin] ERROR - >> key not found: Height or weight must not be zero.`

PR #2101, opened 2024-12-04
OP-1345 | Inpatient Montly Report truncates the year value in the header

ArnaudFofou:
See OP-1345.


dbmalkovsky (approved):
Successfully ran the report with the latest develop and the year is no longer truncated.

PR #2102, opened 2024-12-04
OP-1346 Fix error on hmis108_adm_by_diagnosisOut report

ArnaudFofou:
See OP-1346.


dbmalkovsky (approved):
Ran the report with the latest develop branch and the month value is now present.

PR #2103, opened 2024-12-05
 OP-1348 | Fix error on Laboratory report

ArnaudFofou:
See OP-1348.
Paired with https://github.com/informatici/openhospital-core/pull/1468.


dbmalkovsky (commented):
The table does now print but the second half of the issue (using the Print Label button) still fails.

dbmalkovsky:
Upon more testing and system setup I can get the printer dialog to appear.    So really the problem is when there is no printer just throwing an exception is wrong.  It should be caught and the user told about it without having to dig through the logs.

ArnaudFofou:
> The table does now print but the second half of the issue (using the Print Label button) still fails.

please can you show the exception you have because on my side I don't have exception when I use the Print Label Button

dbmalkovsky:
Disconnect your printer from you machine and try again.    As I have stated previously the exception happens if there is no printer attached.

ArnaudFofou:
> Disconnect your printer from you machine and try again. As I have stated previously the exception happens if there is no printer attached.

I don't have printer connect to my machine but when I click to Print Label I have this window
win

dbmalkovsky:
On Linux before I had any printer  connected I got the exception.

mwithi:
> > Disconnect your printer from you machine and try again. As I have stated previously the exception happens if there is no printer attached.
> 
> I don't have printer connect to my machine but when I click to Print Label I have this window

@ArnaudFofou I think you should try by temporarily removing the favourite printer

ArnaudFofou:
> > > Disconnect your printer from you machine and try again. As I have stated previously the exception happens if there is no printer attached.
> > 
> > 
> > I don't have printer connect to my machine but when I click to Print Label I have this window
> 
> @ArnaudFofou I think you should try by temporarily removing the favourite printer

please How to do it ?

mwithi:
> please How to do it ?

First you may try to turn off Manage Default Printer off ([link](https://www.majorgeeks.com/content/page/manage_default_printer.html))
Then you can try to install a fake printer and set it as default, to simulate @dbmalkovsky exception (I didn't try though)

ArnaudFofou:
> > please How to do it ?
> 
> First you may try to turn off Manage Default Printer off ([link](https://www.majorgeeks.com/content/page/manage_default_printer.html)) Then you can try to install a fake printer and set it as default, to simulate @dbmalkovsky exception (I didn't try though)

I am not able to reproduise that I have turn off Manage Default Printer. And Install fake printer but I don't have and exception after click on the Print Table button

dbmalkovsky:
By looking at the traceback you know where to put the try-catch block and what exception to look for.
You could then write a junit test to simulate it by using mock objects to return the fact the printer is missing.

bundle/language_en.properties
```diff
@@ -706,6 +706,7 @@ angal.lab.tissues.txt
 angal.lab.undefined.txt                                                                                = Undefined
 angal.lab.unknownprocedure.msg                                                                         = Unknown procedure.
 angal.lab.urine.txt                                                                                    = Urine
+angal.lab.noprinter.msg                                                                                = No Printer Available
```
ArnaudFofou:
```suggestion
angal.lab.noprinter.msg                                                                                = No Printer Available.
```

bundle/language_en.properties
```diff
@@ -706,6 +706,7 @@ angal.lab.tissues.txt
 angal.lab.undefined.txt                                                                                = Undefined
 angal.lab.unknownprocedure.msg                                                                         = Unknown procedure.
 angal.lab.urine.txt                                                                                    = Urine
+angal.lab.noprinter.msg                                                                                = No Printer Available.
```
ArnaudFofou:
```suggestion
angal.lab.noprinter.msg                                                                                = No printer available.
```

mwithi:
@ArnaudFofou please update with develop in order to make checks pass

mwithi (approved):
Found that to simulate the "No printer available" on Windows is enough to stop the Print Spoiler service:

image

Something similar should be possible on Linux too.

Out of this, LGTM!

mwithi:
> @ArnaudFofou please update with develop in order to make checks pass

Not needed, was enough to rerun the checks.

PR #2104, opened 2024-12-06
OP-749 | To extend laboratory filter panel with search by patient

ra247362:
See [OH-749](https://openhospital.atlassian.net/jira/software/c/projects/OP/issues/OP-749)

We added a text field in the laboratory section to allow for search by code, as described in the issue. We added error messages when invalid codes were added. The implemented behavior follows the one described in the issue, with the ENTER key being used to search only by code and the SEARCH button used to search by code and date.

Changes were made only in "LabBrowser.java" and the bundled language file "language_en.properties".


src/main/java/org/isf/lab/gui/LabBrowser.java
```diff
@@ -473,14 +514,55 @@ class LabBrowsingModel extends DefaultTableModel {
...
+        public LabBrowsingModel(String exam, LocalDate dateFrom, LocalDate dateTo, String patid) {
+			boolean exists = true;
+            try {
+                Patient pat;
+                if (patid.isEmpty()) {
+                    pat = null;
+                } else {
+                    try {
+                        pat = patManager.getPatientById(Integer.parseInt(patid));
+						exists = pat == null ? false : true;
```
dbmalkovsky:
```suggestion
						exists = pat != null;
```
This is much simpler and does the same thing.  The variable `exists` appears to be used only one in line 532.  Why not just test the variable `pat` there?    And if there is an exception (catch block at line 527) the variable `exists` will not be reset and will use the assigned value from line 518 which is `true`.   So, if there is an exception then line 533 will be executed which given that `pat` is `null` will most likely cause other errors.   It looks like some rethinking of this logic is in order.

bundle/language_en.properties
```diff
@@ -1210,7 +1190,7 @@ angal.newbill.selectamedical.txt
 angal.newbill.selectanexam.txt                                                                         = Select an exam:
 angal.newbill.selectanoperation.txt                                                                    = Select an operation:
 angal.newbill.selectapricelist.title                                                                   = Select a PriceList
-angal.newbill.somepriceshavebeenchangeddoyouwanttoupdatetheitemsprices.fmt.msg                         = Some prices have been changed : {0}\n\nDo you want to update the items' prices?
+angal.newbill.somepriceshavebeenchangeddoyouwanttoupdatetheitemsprices.fmg.msg                         = Some prices have been changed : {0}\n\nDo you want to update the items' prices?
```
dbmalkovsky:
```suggestion
angal.newbill.somepriceshavebeenchangeddoyouwanttoupdatetheitemsprices.fmt.msg                         = Some prices have been changed : {0}\n\nDo you want to update the items' prices?
```
and why did you remove some of the standard suffixes like `.txt` in other places?

src/main/java/org/isf/lab/gui/LabBrowser.java
```diff
@@ -473,14 +514,47 @@ class LabBrowsingModel extends DefaultTableModel {
...
-		public LabBrowsingModel(String exam, LocalDate dateFrom, LocalDate dateTo) {
-			try {
-				pLabs = labManager.getLaboratory(exam, dateFrom.atStartOfDay(), dateTo.atStartOfDay());
-			} catch (OHServiceException e) {
+        public LabBrowsingModel(String exam, LocalDate dateFrom, LocalDate dateTo, String patid) {
+            try {
+				if (!patid.isEmpty()) {
+					Patient pat = patManager.getPatientById(Integer.parseInt(patid));
+					if (pat == null) {
+						pLabs = new ArrayList<>();;
```
dbmalkovsky:
```suggestion
						pLabs = new ArrayList<>();
```

bundle/language_en.properties
ra247362: Hi, we're currently looking into this. As far as we are aware, these changes shouldn't have been made, so we are trying to fix them.

ra247362: We've updated the language bundle to contain our changes, so the issue should be fixed!

bundle/language_en.properties
```diff
@@ -677,6 +677,7 @@ angal.lab.exam
 angal.lab.film.txt                                                                                     = Film
 angal.lab.in                                                                                           = In
 angal.lab.insertvalidage.msg                                                                           = Insert a valid age.
+angal.lab.insertvalidpatientid.msg													= Insert a valid patient ID.
```
dbmalkovsky: Please fix the indentation.

ra247362: Should be fixed now.

mwithi:
Hi @ra247362, to make the checks pass please merge (or rebase) with develop or accept my https://github.com/ra247362/openhospital-gui/pull/1. Thanks!

mwithi (approved):
LGTM. Thanks a lot!

PR #2109, opened 2024-12-20
OP-1304 | Enhance the Code with also to enhance Product Selection

ArnaudFofou:
See OP-1304.


mwithi:
meanwhile the https://github.com/informatici/openhospital-gui/tree/OP-1199-Main-Store-Inventory has changed, so I think the checks fail for that reason, please check

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -2179,129 +2172,43 @@ private boolean areAllMedicalsInInventory() throws OHServiceException {
...
+	    try {
+	        if (!areAllMedicalsInInventory()) {
+	            int userChoice = (!inventoryRowSearchList.isEmpty()) 
+	                ? MessageDialog.yesNo(null, "angal.inventoryrow.doyouwanttoaddallnotyetlistedproducts.msg") 
+	                : JOptionPane.YES_OPTION;
+
+	            if (userChoice == JOptionPane.YES_OPTION) {
+	                jTableInventoryRow.setModel(new InventoryRowModel(medType, nonZero,includeAll, withMovement));
+	                fireInventoryUpdated();
+	                adjustWidth();
```
mwithi: In particular, this method has been removed, because by introducing `jTableInventoryRow.setAutoCreateColumnsFromModel(false)` just after first JTable creation, we don't have to worry about the appearence every time we update the model

ArnaudFofou: you talk about adjustWidth(); rigth? I have already remove it.

mwithi: my comment was before your removal... ok, thanks

mwithi (approved):
Did many tests and LGTM. Thanks!

PR #2115, opened 2025-01-09
OP-1376 OP-1377 |  Fix some Bug

ArnaudFofou:
See OP-1376 and OP-1377.

- [x] Check If real quantity is greatter than zero during validation.
- [x] Change the default behavior of a JFrame of Lot Information.
- [x]  when chargeCombo, dischargeCombo, supplierCombo, destinationCombo or real Quantity has been updated, ask the user to update Inventory before validate it.
- [x] avoid to close the windows on `edit mode` when the user has clean table and  close the window directly.


bundle/language_en.properties
```diff
@@ -724,6 +726,7 @@ angal.inventory.doyoureallywanttoconfirmthisinventory.msg
 angal.inventoryrow.notfound.msg                                                                        = Inventory not found.
 angal.inventoryrow.invalidquantity.msg                                                                 = Invalid quantity.
 angal.inventory.pleaseinsertavalidinventorydate.msg                                                    = Please insert a valid inventory date.
+angal.inventoryrow.pleaseinsertatleastoneinventoryrow.msg                                              = Please insert at least one inventory row and save \n or reset the field before close the window.
```
dbmalkovsky:
```suggestion
angal.inventoryrow.pleaseinsertatleastoneinventoryrow.msg                                              = Please insert at least one inventory row and save \n or reset the field before closing the window.
```

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
```diff
@@ -909,13 +920,13 @@ private JButton getCloseButton() {
...
+				int reset = MessageDialog.yesNo(null, "angal.inventoryrow.doyouwanttosavethechanges.msg");
 				if (reset == JOptionPane.YES_OPTION) {
 					this.saveButton.doClick();
-				} else {
-					resetVariable();
-					dispose();
 				}
+			}
+			if (!mode.equals("new") && inventoryRowSearchList.isEmpty()) {
+				MessageDialog.info(null, "angal.inventoryrow.pleaseinsertatleastoneinventoryrow.msg");
```
dbmalkovsky: Why isn't this an `error`?  Some action by the user is required; I don't expect that for `info`

ArnaudFofou: It can be and error or and info. but for me info is more better because the user clean the table and then try to close the window without enter at least one inventory row. but I have update to set as error. If you agree I can revert to info

dbmalkovsky: You are telling the user to "please insert at least one inventory row" .... sure seems like they have committed an error that you want them to fix. IMHO.  But do what you want .....

bundle/language_en.properties
```diff
@@ -670,6 +670,8 @@ angal.hospital.visitduration.txt
 angal.hospital.visitendhour.txt                                                                        = Visit end hour
 angal.hospital.visitstarthour.txt                                                                      = Visit start hour
 angal.inventory.addproduct.label                                                                       = Add products
+angal.inventory.allinventoryrowshouldhaverealqtygreatterthanzero.msg                                   = All inventory(ies) row(s) should have real quantity greater than zero.
```
mwithi: This is not true.... I try to explain again but please check that the whole code is consistent:

If the lot is NOT NEW, the real quantity CAN be zero, because we are doing an inventory and we are detecting that one existing product/lot is missing.
If the lot is NEW, the real quantity must be GREATER than zero, because it is not possibile to detect a new product/lot if it is missing.

In this extend, the message is not correct. But please, check the logic that uses this message.

src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
ArnaudFofou: Okay

bundle/language_en.properties
ArnaudFofou: I have update the message

mwithi: did you verify the logic if it **really** checks that only for new lots?

ArnaudFofou: yes look here: https://github.com/informatici/openhospital-gui/pull/2115/files#diff-6f7e2e3cd7655423fb3b6f6a9c6956e0118bf570efb584289a9f21cfb6fac3c7R1021

mwithi (approved):
LGTM. Thanks a lot!

PR #2117, opened 2025-01-14
OP-1306 Fix alcool -> alcohol

mwithi:
See OP-1306.
Paired with https://github.com/informatici/openhospital-core/pull/1470.


dbmalkovsky (approved):
LGTM

PR #2121, opened 2025-01-15
OP-1379 Fix lot creation for rectify in ward pharmacy

mwithi:
See OP-1379.

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
@@ -538,7 +538,7 @@ public class WardPharmacyRectify extends JDialog {
 		JTextField lotNameTextField = new JTextField(15);
 		lotNameTextField.addAncestorListener(new RequestFocusListener());
 		TextPrompt suggestion = new TextPrompt(MessageBundle.getMessage("angal.medicalstock.multiplecharging.lotid"), lotNameTextField);
-		suggestion.setFont(new Font("Tahoma", Font.PLAIN, 14)); //$NON-NLS-1$
+		suggestion.setFont(FONT_PLAIN_14);
 		suggestion.setForeground(Color.GRAY);
 		suggestion.setHorizontalAlignment(SwingConstants.CENTER);
 		suggestion.changeAlpha(0.5f);
@@ -615,7 +615,6 @@ public class WardPharmacyRectify extends JDialog {
 
 	protected int askQuantity(Medical med) {
 		StringBuilder title = new StringBuilder(MessageBundle.getMessage("angal.common.quantity.txt"));
-		StringBuilder message = new StringBuilder(med.toString());
 		String prodCode = med.getProdCode();
 		if (prodCode != null && !prodCode.equals("")) {
 			title.append(' ').append(MessageBundle.getMessage("angal.common.code.txt")); //$NON-NLS-1$ //$NON-NLS-2$
@@ -626,7 +625,7 @@ public class WardPharmacyRectify extends JDialog {
 		int qty = 0;
 		do {
 			String quantity = JOptionPane.showInputDialog(this,
-				message.toString(),
+				med.toString(),
 				title.toString(),
 				JOptionPane.QUESTION_MESSAGE);
 			if (quantity != null) {
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
@@ -34,6 +34,7 @@ import java.awt.GridBagLayout;
 import java.awt.GridLayout;
 import java.awt.Insets;
 import java.math.BigDecimal;
+import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.EventListener;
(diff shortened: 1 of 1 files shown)


src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -617,9 +626,9 @@ protected int askQuantity(Medical med) {
 		int qty = 0;
 		do {
 			String quantity = JOptionPane.showInputDialog(this,
-							message.toString(),
-							title.toString(),
-							JOptionPane.QUESTION_MESSAGE);
+				message.toString(),
```
dbmalkovsky: `message` is defined at line 618 as:
<pre>
StringBuilder message = new StringBuilder(med.toString());
</pre>

Then why not change this line this and remove line 618 as this is the only two places `message` is used:
```suggestion
				med.toString(),
```

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -528,28 +529,24 @@ private MedicalWard chooseLot(Medical medical, boolean newLot) {
...
+		LocalDateTime preparationDate;
+		LocalDateTime expiringDate;
 		Lot lot = null;
-		JTextField lotNameTextField = new JTextField(15);
 
+		JTextField lotNameTextField = new JTextField(15);
+		lotNameTextField.addAncestorListener(new RequestFocusListener());
 		TextPrompt suggestion = new TextPrompt(MessageBundle.getMessage("angal.medicalstock.multiplecharging.lotid"), lotNameTextField);
-		suggestion.setFont(FONT_PLAIN_14);
+		suggestion.setFont(new Font("Tahoma", Font.PLAIN, 14)); //$NON-NLS-1$
```
dbmalkovsky: Curious as to why you only changed this CONSTANT to  its expansion and then didn't remove the definition at line 88?

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
mwithi: I guess it was there since long time preparing for a longer message... Let me remove that.

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -560,20 +557,32 @@ protected Lot askLot() {
...
-				if (expireDateChooser.getDate().isBefore(preparationDateChooser.getDate())) {
-					MessageDialog.error(this, "angal.medicalstockward.rectify.expirydatebeforepreparationdate");
-				} else {
-					LocalDateTime expiringDate = expireDateChooser.getDateEndOfDay();
-					LocalDateTime preparationDate = preparationDateChooser.getDateStartOfDay();
-					lot = new Lot(lotName, preparationDate, expiringDate);
+				try {
+					if (lotName.isEmpty()) {
+						MessageDialog.error(this, "angal.medicalstockward.rectify.lotnumberSelect");
+						continue;
```
dbmalkovsky, line 568: why the `continue`?  This isn't a loop and it will fall through to the next `if` statement anyway....

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
```diff
@@ -560,20 +557,32 @@ protected Lot askLot() {
...
+					}
+					if (movStockInsertingManager.lotExists(lotName)) {
+						MessageDialog.error(this, "angal.medicalstock.multiplecharging.theinsertedlotcodealreaedyexists.msg");
+						continue;
```
dbmalkovsky, line 572: same as line 568

dbmalkovsky (commented):
Left some comments that need looking into ....

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
mwithi: just applying most recent code from MovStockMultipleCharging, apparently this one was more optimized... I will rollback this. Thanks!

src/main/java/org/isf/medicalstockward/gui/WardPharmacyRectify.java
mwithi: Are you sure? The continue statement in a loop immediately skips the rest of the current iteration and moves control to the next iteration of the loop... to proceed to the next `if` there would be no need for anything

mwithi: sorry, the loop is `do ... while`

dbmalkovsky: I didn't look far enough back; sorry.

But I am scratching my head to see why it is `do ... while`  seems strange; the logic needs to be re-examined at some point.

mwithi: It keeps asking the lot information until a lot is chosen or a `return` closes the dialog

mwithi: hello @dbmalkovsky is it OK the explanation? If so, can we resolve the two comments and approve? Thanks!

dbmalkovsky (approved):
LGTM

PR #2122, opened 2025-01-15
OP-1378 Fix Alvo -> Bowel

mwithi:
See OP-1378.
Paired with https://github.com/informatici/openhospital-core/pull/1472.


dbmalkovsky (approved):
LGTM

PR #2131, opened 2025-02-18
OP-1380: replace deprecated method

dbmalkovsky:
See OP-1380

Matched with https://github.com/informatici/openhospital-core/pull/1479


mwithi (approved):
LGTM. Thanks a lot!

PR #2140, opened 2025-03-21
Chore: add LOG_DEST directory to .gitignore

dbmalkovsky:
Just trying to clean up my local changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ logs
 /rsc/*.pref
 /rsc/*.xml
 /target/
+LOG_DEST/


.gitignore
```diff
@@ -15,3 +15,4 @@ logs
 /rsc/*.pref
 /rsc/*.xml
 /target/
+LOG_DEST/
```
mwithi, line 18: this should be a dev local folder and could be anything, I don't know if to add to .gitignore because of course should not be committed (whatever it is)

dbmalkovsky:
I'll look into this more later.

PR #2154, opened 2025-09-18
Fix: log4j/log4j2 mixed syntax

mwithi:
In **Log4j 2** with _.properties_ files, we shouldn't use keys that begin with `log4j.logger` or `log4j.additivity`:

- That's old Log4j 1.x syntax.
- Log4j 2 instead expects the form `logger.<name>.*`

The same goes for `log4j.formatMsgNoLookups=true`: 

- That property is now set via system property (-Dlog4j2.formatMsgNoLookups=true) 
- or in `log4j2.component.properties` (added this file)

OR we can just get rid of the `log4j2.formatMsgNoLookups` and leave to the SA to decide if to manage it or not.

### DIFF
--- a/rsc/.gitignore
+++ b/rsc/.gitignore
@@ -1,1 +1,10 @@
-/generalData.properties
+database.properties
+dicom.properties
+examination.properties
+log4j2-spring.properties
+settings.properties
+sms.properties
+telemetry.properties
+txtPrinter.properties
+xmpp.properties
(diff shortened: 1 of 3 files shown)


dbmalkovsky (approved):
LGTM

mizzioisf (approved):
LGTM also.
Really sorry for the delay :-/

PR #2174, opened 2026-01-15
Fix for log file path

mizzioisf:
Fix for log file path management: solves the bug that creates the file:

[$OH_PATH]/data/log/openhospital.log/openhospital.log (wrong)

instead of the:

[$OH_PATH]/data/log/openhospital.log (correct)


dbmalkovsky (dismissed):
LGTM

PR #2177, opened 2026-01-20
Menu Expert fix

mizzioisf:
Fix for script menu: Expert mode it is shown only if "E" option is selected
Some commented, old code has also been removed.


mwithi (approved):
LGTM

PR #2179, opened 2026-01-23
OP-1398 Fix DB datetime error when using Arabic locale

mwithi:
See OP-1398.


src/main/java/org/isf/menu/gui/MainMenu.java
```diff
@@ -153,6 +153,7 @@ public MainMenu(User myUserIn) {
 		GeneralData.initialize();
 		this.activableModules = retrieveActivatedModulesMap();
 		Locale.setDefault(new Locale(GeneralData.LANGUAGE)); // for all fixed options YES_NO_CANCEL in dialogs
+		Locale.setDefault(Locale.Category.FORMAT, Locale.ROOT); // fix technical formats after SpringBoot 3.4.5
```
dbmalkovsky, line 156: why isn't line 155 removed as it seems to be superseded by line 156?

mwithi: both are needed... apparently the setDefault is the easiest way to set all Swing components (instead to tamper directly with them one by one). Line 156 is to undo the setDefault on the FORMAT category only.

dbmalkovsky (approved):
LGTM

PR #2193, opened 2026-03-17
OP-1401 Add support for direct TXT printing for thermal POS printers

mwithi:
See OP-1401.
Paired with https://github.com/informatici/openhospital-core/pull/1564


dbmalkovsky (approved):
LGTM

PR #2194, opened 2026-04-14
OP-1405 Fix bill amount and bill balance recomputation upon save

mwithi:
See OP-1405.
Paired with https://github.com/informatici/openhospital-core/pull/1578.


dbmalkovsky (approved):
LGTM with light testing

PR #2196, opened 2026-05-17
OP-214: Restructure report properties into language subfolders

4bh1gn4:
See OP-214.

## Summary
Restructures language-specific .properties files from flat naming (PatientBill_en.properties) into language subfolders (en/PatientBill.properties) across rpt_base, rpt_stat, and rpt_extra.

## Changes
- Migrated all language-specific .properties files into language subfolders
- Default .properties files remain in the root alongside .jrxml files
- Updated ReportLauncher.java to resolve language-specific paths

Tested locally in English, Spanish, and French - reports load correctly from language subfolders

#### Depends on openhospital-core PR
Link: https://github.com/informatici/openhospital-core/pull/1592


rpt_base/en/PatientBill.properties
mwithi, line 1: Hi @4bh1gn4. I've noticed that the <ReportName>.properties (default) file must be aside the <ReportName>.jrxml one, for a limitation of Jasper Studio that uses the java standard pattern for bundles. This for the sake of reports design.

So the final structure should something like:
rpt_base/PatientBill.jrxml
rpt_base/PatientBill.jasper
rpt_base/PatientBill.properties (default needed by Jasper Studio)
rpt_base/en/report.properties (preferred structure for Transifex)

oh.bat
```diff
@@ -326,7 +326,7 @@ for %%A IN (%OH_PATH%\%OH_DIR%\lib\*.jar) DO (
 	set CLASSPATH=!CLASSPATH!;%%A
 )
 set CLASSPATH=%CLASSPATH%;%OH_PATH%\%OH_DIR%\bundle
-set CLASSPATH=%CLASSPATH%;%OH_PATH%\%OH_DIR%\rpt_base
+set CLASSPATH=%CLASSPATH%;%OH_PATH%\%OH_DIR%\rpt_base\%OH_LANGUAGE%
 set CLASSPATH=%CLASSPATH%;%OH_PATH%\%OH_DIR%\rpt_extra
 set CLASSPATH=%CLASSPATH%;%OH_PATH%\%OH_DIR%\rpt_stat
```
mwithi, line 331: The problem comes here, because if we put both `rpt_base` and `rpt_base/<lang>` in the classpath, the two properties with the same filename would conflict. Maybe we have to fetch properties files directly from filesystem instead of classpath... could you try this and let me know? Thanks!

mwithi: Moreover, we already use to override jasper built-in parameter REPORT_RESOURCE_BUNDLE to pass the bundle we want (skipping the JR java default pattern). In this way we are also able to fallback to the default properties if the lang specific one is missing (rather than have an exception)

mwithi (commented):
Hi @4bh1gn4 very good job, thanks! Nevertheless, I would consider leaving the default properties aside the jrxml file for the sake of desing report within Jasper Studio. From there, more considerations follow...

mwithi (approved):
LGTM. Thanks a lot!

PR #2202, opened 2026-05-29
OP-1383 Fix XMPP i18n key typo and align language bundles

NgKokHua:
**Summary**
This PR fixes a typo in an XMPP translation key and aligns code and bundle keys to use the same canonical key name across locales.

**What changed**
Updated the XMPP notification key usage in the GUI code to the correctly spelled key.
Removed/replaced the misspelled key variant in language bundles.
Kept the message content unchanged; only key naming consistency was addressed.

**Why**
The previous typo created two near-identical keys and risked inconsistent usage over time.
Standardizing on one correct key prevents future drift and makes i18n maintenance safer.

**Scope**
Focused key normalization only.
No functional workflow changes beyond using the corrected key.
This is not a full orphan-key cleanup pass.

**Validation**
Verified the corrected key is referenced in code.
Verified the old, misspelled key is no longer referenced.
Confirmed locale bundles are aligned to the corrected key name.

**Risk**
Low.
Change is limited to i18n key identifiers and does not alter business logic.


mwithi (approved):
LGTM. Thanks a lot!

PR #2209, opened 2026-06-16
OP-219 Let the user load a DICOM image into an existing series

giuseppesorge:
## What
When loading an image in the DICOM viewer, the user can now choose to load it into an **already existing** series of the patient, instead of always creating a new one.

Resolves [OP-219](https://openhospital.atlassian.net/browse/OP-219).

## How
- In `DicomGui`, after the pre-load dialog (where the `//TODO` used to be), if the patient already has at least one series, a small dialog asks whether to create a **New series** or load into an existing one. The series currently selected in the thumbnail panel is pre-selected.
- When an existing series is chosen, its identifiers (series number, instance UID, UID, description) are copied onto the file so it joins that series.
- Added the i18n keys (`angal.dicom.newseries.txt`, `angal.dicom.selectseries.title`, `angal.dicom.selectserieswheretoloadthefiles.msg`) to the English bundle.

## Depends on
informatici/openhospital-core#1594 (same branch name) — the core honors the preselected series for actual DICOM files. The CI builds this PR against that core branch automatically.

## Testing
- `mvn package` (CI gate) green; core DICOM tests pass.
- Verified end-to-end against a running instance with two real `.dcm` files: picking the existing series loads both images into one series (confirmed in the viewer and in the DB: a single `DM_FILE_SER_INST_UID` with 2 rows).

## Note
The user-manual screenshot (oh-doc) can be refreshed as a follow-up, per the ticket.

[OP-219]: https://openhospital.atlassian.net/browse/OP-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -509,7 +509,10 @@ angal.dicom.image.zoom                                                          
 angal.dicom.load.btn                                                                                   = Load Image
 angal.dicom.load.btn.key                                                                               = L
 angal.dicom.loading                                                                                    = Loading...
+angal.dicom.newseries.txt                                                                              = New series
 angal.dicom.open.txt                                                                                   = Open Image
+angal.dicom.selectseries.title                                                                         = Select series
+angal.dicom.selectserieswheretoloadthefiles.msg                                                        = Select the series to load the file(s) into:
 angal.dicom.thefileisinanunknownformat.fmt.msg                                                         = The file is in an unknown format: {0}.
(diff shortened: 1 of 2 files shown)


mwithi (approved):
LGTM. Thanks a lot!

PR #2210, opened 2026-06-17
OP-1388 Ask for an optional reason when deleting a stock movement

giuseppesorge:
## OP-1388 — Optional reason when deleting a stock movement

The Stock Movement Browser and Ward Pharmacy delete actions now prompt for an optional reason, which is passed to the core so the deletion is recorded in the audit log (see the paired core PR).

### Bundled bug fix
Also fixes a pre-existing bug: after deleting the last movement, **the deleted row stayed visible in the Stock Movement Browser table until the application was restarted**. The in-place refresh re-runs the Filter action, which threw a `NullPointerException` when the optional *Lot Preparation/Due Date* filters were left empty (`TimeTools.getBeginningOfDay(null)`). Those dates are now handled null-safely, so the row disappears immediately after deletion — and the Filter button no longer fails when the lot date filters are empty.

### Notes
- New i18n key `angal.medicalstock.deletemovementreason.msg` (English bundle).
- Requires the paired core PR informatici/openhospital-core#1595 (same branch name, so CI cross-builds against it).

JIRA: https://openhospital.atlassian.net/browse/OP-1388

### DIFF
--- a/src/main/java/org/isf/medicalstock/gui/MovStockBrowser.java
+++ b/src/main/java/org/isf/medicalstock/gui/MovStockBrowser.java
@@ -892,21 +892,21 @@ public class MovStockBrowser extends ModalJFrame {
 				if (!isAutomaticLot()) {
 					model = new MovBrowserModel(medicalSelected,
 									medicalTypeSelected, wardSelected, movementTypeSelected,
-									TimeTools.getBeginningOfDay(movDateFrom.getDateStartOfDay()),
-									TimeTools.getBeginningOfNextDay(movDateTo.getDateStartOfDay()),
-									TimeTools.getBeginningOfDay(lotPrepFrom.getDateStartOfDay()),
-									TimeTools.getBeginningOfNextDay(lotPrepTo.getDateStartOfDay()),
-									TimeTools.getBeginningOfDay(lotDueFrom.getDateStartOfDay()),
-									TimeTools.getBeginningOfNextDay(lotDueTo.getDateStartOfDay()));
+									beginningOfDayOrNull(movDateFrom.getDateStartOfDay()),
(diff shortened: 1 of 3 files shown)


src/main/java/org/isf/medicalstock/gui/MovStockBrowser.java
```diff
@@ -1026,12 +1034,14 @@ private JButton getDeleteLastMovementButton() {
...
-						movBrowserManager.deleteLastMovement(lastMovement);
-					} else {
+					JTextField reasonField = new JTextField(20);
+					Object[] message = { MessageBundle.getMessage("angal.medicalstock.deletemovementreason.msg"), reasonField };
+					int answer = JOptionPane.showConfirmDialog(this, message,
+						MessageBundle.getMessage("angal.common.delete.btn"), JOptionPane.YES_NO_OPTION, JOptionPane.QUESTION_MESSAGE);
+					if (answer != JOptionPane.YES_OPTION) {
 						return;
 					}
+					movBrowserManager.deleteLastMovement(lastMovement, reasonField.getText());
```
mwithi: I agree with the Yes/No option. Anyway I think that, given the auditing purpose of this change, the reason must be compulsory rather than optional.

giuseppesorge: Done in 9ccca23d — the reason is now mandatory: the dialog re-prompts until a non-empty reason is entered (or you cancel). I applied it to both delete paths (Pharmaceutical Stock here and Pharmaceutical Stock Ward in `WardPharmacy`) for consistency, and reworded the confirmation message. It's enforced in the GUI; I can also enforce it in the manager (`deleteLastMovement`) if you'd prefer it guaranteed at the service layer.

mwithi (approved):
LGTM. Thanks a lot!

PR #2212, opened 2026-06-18
OP-1216 Adapt GUI to DECIMAL/INT money and quantity types

giuseppesorge:
## OP-1216 — Adapt the Swing GUI to the DECIMAL/INT type change

Adapts the GUI to the core type change (OP-1216) and fixes two issues found
while testing the feature end-to-end against a migrated demo database.

### Changes
- Accounting totals (Balance/Payments/User) and `PatientBillEdit` use the
  `BigDecimal` amount/balance getters directly.
- Price model/browser handle `BigDecimal` price (display narrows to `double`
  only for rendering); `MedicalEdit` and the Ward Pharmacy screens build
  movements with `BigDecimal`/`int` quantities.
- **Fix — fractional ward dispensing** (`WardPharmacyNew`): the dispense
  quantity was parsed with `Integer.parseInt`, so a half-unit (e.g. `2.5`) was
  rejected with "Insert a valid value" even though the ward movement quantity is
  DECIMAL. The charging quantity is now handled as `BigDecimal` end-to-end.
  Incoming ward stock stays integer, per the ticket.
- **`WardPharmacyRectify`**: the rectification arithmetic (`lotQty`, `newQty`,
  `movQuantity`) runs on `BigDecimal` and the movement quantity is passed
  through unchanged; the unit-cost prompt parses straight into `BigDecimal`.
  The quantity spinner keeps a `Double`-based model (`SpinnerNumberModel`
  increments a `BigDecimal` value through `longValue()`, which would break the
  0.5 stepping), and every `setValue` call is normalized to `Double`.
- **Inventory screens**: `InventoryWardEdit` and `InventoryEdit` adapted to the
  `BigDecimal` `MedicalInventoryRow`. The ward inventory accepts fractional
  real quantities (quantity columns use `BigDecimal` and the existing decimal
  renderer); the main-store inventory keeps integer entry and rejects
  fractional input with the standard invalid-quantity error, consistent with
  the integer main-store stock.
- **Fix — Price Lists screen crash** (`PriceModel`): `getValueAt` called
  `price.getPrice().doubleValue()`, but group/category nodes in the price tree
  have a null price, so the whole Price Lists screen rendered an empty tree
  (repeated `NullPointerException`). Guarded the null case.
- Tests updated for the new types.

### Verification
- `mvn package` green.
- Manual end-to-end on a migrated demo DB: a bill with items `0.1 + 0.2` totals
  exactly `0.3`; editing a price to `19.99` persists exactly; dispensing `2.5`
  from a ward stores the movement and out-quantity as `2.50`; a ward inventory
  row saved with real quantity `1.5` persists as an exact `1.50`; a ward
  rectification of `2.5` stores an exact `-2.50` movement; typing `2.5` in the
  main-store inventory is rejected, `3` is accepted.

Requires the core change. Companion PRs (same branch name for CI cross-build)
are linked in a comment below.

JIRA: https://openhospital.atlassian.net/browse/OP-1216

### DIFF
--- a/src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
+++ b/src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
@@ -162,7 +162,7 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 	private Patient patientSelected;
 	private float patientWeight;
 	private Ward wardSelected;
-	private Object[] medClasses = { Medical.class, Double.class, String.class };
+	private Object[] medClasses = { Medical.class, BigDecimal.class, String.class };
 	private String[] medColumnNames = {
 			MessageBundle.getMessage("angal.wardpharmacy.medical.col").toUpperCase(),
 			MessageBundle.getMessage("angal.common.quantity.txt").toUpperCase(),
@@ -323,16 +323,16 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 		}
 	}
 
-	private boolean checkQuantityInLot(MedicalWard medWard, double qty) {
-		double wardQty = medWard.getQty();
-		if (qty > wardQty) {
+	private boolean checkQuantityInLot(MedicalWard medWard, BigDecimal qty) {
+		BigDecimal wardQty = medWard.getQty();
+		if (qty.compareTo(wardQty) > 0) {
 			MessageDialog.error(this, "angal.medicalstock.movementquantityisgreaterthanthequantityof.fmt.msg", qty, wardQty);
 			return false;
 		}
 		return true;
 	}
 
-	private MedicalWard automaticChoose(List<MedicalWard> drug, String me, double quantity) {
+	private MedicalWard automaticChoose(List<MedicalWard> drug, String me, BigDecimal quantity) {
 		drug.sort((o1, o2) -> {
 			if (o1.getLot().getDueDate() == null || o2.getLot().getDueDate() == null) {
 				return 0;
@@ -341,15 +341,15 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 		});
 
 		MedicalWard medWard = null;
-		double q = quantity;
+		BigDecimal q = quantity;
 		for (MedicalWard elem : drug) {
 			if (elem.getMedical().getDescription().equals(me)) {
 
-				if (elem.getQty() != 0.0) {
-					if (q > 0) {
-						if (elem.getQty() <= q) {
-							double maxquantity = elem.getQty();
-							q = q - maxquantity;
+				if (elem.getQty().signum() != 0) {
+					if (q.signum() > 0) {
+						if (elem.getQty().compareTo(q) <= 0) {
+							BigDecimal maxquantity = elem.getQty();
+							q = q.subtract(maxquantity);
 							medWard = elem;
 							addItem(medWard, maxquantity);
 
@@ -357,7 +357,7 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 							medWard = elem;
 
 							addItem(medWard, q);
-							q = 0;
+							q = BigDecimal.ZERO;
 						}
 					}
 
@@ -368,12 +368,12 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 		return medWard;
 	}
 
-	private MedicalWard chooseLot(List<MedicalWard> drug, String me, double quantity) {
+	private MedicalWard chooseLot(List<MedicalWard> drug, String me, BigDecimal quantity) {
 		List<MedicalWard> dr = new ArrayList<>();
 		MedicalWard medWard = null;
 		for (MedicalWard elem : drug) {
 			if (elem.getMedical().getDescription().equals(me)) {
-				if (elem.getQty() != 0.0) {
+				if (elem.getQty().signum() != 0) {
 					MedicalWard e = elem;
 					dr.add(e);
 				}
@@ -413,14 +413,14 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 		return medWard;
 	}
 
-	protected double askQuantity(String med, List<MedicalWard> drug) {
-		double qty = 0;
-		double totalQty = 0;
+	protected BigDecimal askQuantity(String med, List<MedicalWard> drug) {
+		BigDecimal qty = BigDecimal.ZERO;
+		BigDecimal totalQty = BigDecimal.ZERO;
 		String prodCode = null;
 		for (MedicalWard elem : drug) {
 
 			if (med.equals(elem.getMedical().getDescription())) {
-				totalQty += elem.getQty();
+				totalQty = totalQty.add(elem.getQty());
 				prodCode = elem.getMedical().getProdCode();
 			}
 
@@ -446,17 +446,17 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 
 			if (quantity != null) {
 				try {
-					qty = Double.parseDouble(quantity);
-					if (qty == 0) {
-						return 0;
+					qty = new BigDecimal(quantity.trim());
+					if (qty.signum() == 0) {
+						return BigDecimal.ZERO;
 					}
-					if (qty < 0) {
+					if (qty.signum() < 0) {
 						throw new NumberFormatException();
 					}
 
 				} catch (NumberFormatException nfe) {
 					MessageDialog.error(this, "angal.medicalstock.multipledischarging.pleaseinsertavalidvalue");
-					qty = 0;
+					qty = BigDecimal.ZERO;
 				}
 			} else {
 				return qty;
@@ -471,15 +471,15 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 			} else {
 				askQuantity(med, wardDrugs);
 			}
-		} while (qty == 0);
+		} while (qty.signum() == 0);
 
 		return qty;
 
 	}
 
-	private boolean checkQuantity(double totalQty, double qty) {
+	private boolean checkQuantity(BigDecimal totalQty, BigDecimal qty) {
 
-		if (qty > totalQty) {
+		if (qty.compareTo(totalQty) > 0) {
 			StringBuilder message = new StringBuilder();
 			message.append(MessageBundle.getMessage("angal.medicalstock.multipledischarging.thequantityisnotavailable")) //$NON-NLS-1$
 				.append('\n') // $NON-NLS-1$
@@ -498,7 +498,7 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 			jButtonAddMedical.setIcon(new ImageIcon("rsc/icons/plus_button.png")); //$NON-NLS-1$
 			jButtonAddMedical.addActionListener(actionEvent -> {
 				String medical = (String) jComboBoxMedicals.getSelectedItem();
-				double quantity = askQuantity(medical, wardDrugs);
+				BigDecimal quantity = askQuantity(medical, wardDrugs);
 			});
 		}
 		return jButtonAddMedical;
@@ -524,7 +524,7 @@ public class WardPharmacyNew extends JDialog implements SelectionListener {
 		return jButtonRemoveMedical;
 	}
 
-	private void addItem(MedicalWard ward, double quantity) {
+	private void addItem(MedicalWard ward, BigDecimal quantity) {
 		if (ward != null) {
 
 			MedicalWard item = new MedicalWard(ward.getMedical(), quantity, ward.getId().getLot());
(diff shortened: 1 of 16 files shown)


giuseppesorge:
OP-1216 spans three repositories (same branch name `OP-1216-decimal-precision` so the gui/api CI cross-builds the core fork branch):
- core: https://github.com/informatici/openhospital-core/pull/1597
- api: https://github.com/informatici/openhospital-api/pull/586
- gui: https://github.com/informatici/openhospital-gui/pull/2212

giuseppesorge:
Heads-up on merge order: this PR depends on the core change in informatici/openhospital-core#1597 (money/quantity converted to DECIMAL/INT, i.e. `BigDecimal`). CI here is red only because the gui build compiles against the **develop** `openhospital-core` SNAPSHOT, which still exposes `Double` — so the adapted code (e.g. `PatientBillEdit`) reports `BigDecimal`/`Double` mismatches. Verified locally that this branch **compiles cleanly (BUILD SUCCESS) and tests pass against core #1597**. So please merge core #1597 first (and api #586 alongside); CI here should then go green. Happy to rebase once core lands.

src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
mwithi, line 1: Not possibile to replace double with BigDecimal also in this class?

giuseppesorge: The dispensed quantity is already persisted exactly: the movement is stored as `new BigDecimal(Double.toString(qty))`, which round-trips any decimal the user types without floating-point error. Converting the intermediate arithmetic in this class alone wouldn't add precision, because `MedicalWard` carries the ward-stock quantity as a transient `Double` (`getQty()`/`setQty()`), so the value would round-trip through `double` again at `addItem()`/`getQty()`. A genuine end-to-end `BigDecimal` would mean migrating `MedicalWard.getQty()`/`setQty()` (and the `MedicalWard(Medical, double, Lot)` constructor) in core, which touches many call sites — I'd suggest doing that as a dedicated follow-up rather than in this PR. Happy to open the ticket if you agree.

mwithi: Alright, just to not forget this point, thanks!

giuseppesorge: Tracked as OP-1422 ("Use BigDecimal end-to-end for ward stock quantities"), linked to OP-1216 — so `MDSRWRD_IN_QTI` / `MedicalWard.qty` and the ward-transfer path get the full BigDecimal treatment there. Thanks!

mwithi: Now that we clarified the meaning of my previous comment, I was referring to preferably use BigDecimal for parsing and internal quantity arithmetic, such as:

```
double q = quantity;
q = q - maxquantity;
...
double totalQty = 0;
totalQty += elem.getQty();
```

that doubel Qty derives from:

```
public class MedicalWard

	...
	@Transient
	private Double qty = 0.0;

```

but the point is not to just wrapping the final value into BigDecimal, but to avoid calculation in floating points from now on.
Hence my comment was to fix that part too.

mwithi:
Hi @giuseppesorge, sorry for the mess, I was trying to fix the CI because it seems that some GITHUB variables changed. 
https://github.com/informatici/openhospital-gui/pull/2212/commits/4c91bab972925db8bbaaed7ad02dced927db975f was a fix by the way.

giuseppesorge:
@mwithi done — `WardPharmacyNew` now parses and computes the ward-movement quantities in BigDecimal rather than double/int: the dispense quantity is parsed with `new BigDecimal(input)`, and the running totals, availability checks and lot splitting use `compareTo`/`signum`/`subtract`, so nothing goes through floating point. To make that possible the transient `MedicalWard.qty` aggregate is now `BigDecimal` (core #1597); `WardPharmacy` and `WardPharmacyRectify` are adapted to it.

Scope note: `MedicalInventoryManager` and the Jasper print DTO `MedicalWardForPrint` keep their own double-based models and convert at the boundary, so no inventory/report behaviour changes. The spinner-based math in `WardPharmacyRectify` (lot rectify) is left as-is since it's driven by a JSpinner, not the point you raised — happy to convert that too if you'd prefer.

Verified end-to-end on a demo DB: a 2.5 dispense to a patient is stored exactly (ward OUT +2.50) and a 2.5 ward-to-ward transfer is rejected as non-integer, while an integer transfer moves the stock correctly.

mwithi:
@giuseppesorge thanks! Please do the same whenever calculations concern prices and quantities:

A few remaining places look worth aligning:

1. In core, the ward-stock aggregate queries still seem to return `Double` (`findQuantityInWardWhereMedical` and `findQuantityInWardWhereMedicalAndWard`), and some callers then cast/truncate the result to `int`. Since `MDSRWRD_OUT_QTI` is now `DECIMAL`, the available ward quantity can legitimately be fractional after patient dispensing, so `sum(medWard.in_quantity-medWard.out_quantity)` should probably return/use `BigDecimal` end-to-end. Truncation should happen only where the domain explicitly requires whole units, such as ward-to-ward incoming stock validation.

2. `WardPharmacyRectify`, `MedicalInventoryManager`,  ok to adapt

3. Some core tests still use `isCloseTo` after converting values to `doubleValue()`.

4. Minor: the GUI PR description still says the ward charging flow was switched to `double`, but the latest code now uses `BigDecimal` in `WardPharmacyNew`. It would be good to update the description to avoid confusing reviewers.

Thanks!

giuseppesorge:
Good point. The deeper change — using `BigDecimal` for the internal quantity arithmetic (`MedicalWard.qty`, the `q`/`totalQty` computations in `WardPharmacyNew`), not just boxing at the boundary — is tracked in OP-1422. Per Alessandro's note there, it will be scoped to ward OUT-movement quantities only (fractional dispensing to patients), keeping ward IN qty and ward-to-ward transfers integer. Keeping #2212 to the type adaptation so it can land.

mwithi:
> Good point. The deeper change — using `BigDecimal` for the internal quantity arithmetic (`MedicalWard.qty`, the `q`/`totalQty` computations in `WardPharmacyNew`), not just boxing at the boundary — is tracked in [OP-1422](https://openhospital.atlassian.net/browse/OP-1422). Per Alessandro's note there, it will be scoped to ward OUT-movement quantities only (fractional dispensing to patients), keeping ward IN qty and ward-to-ward transfers integer. Keeping #2212 to the type adaptation so it can land.

OP-1422 is canceled now, can you apply the needed changes within the OP-1216 scope? Thanks.

[OP-1422]: https://openhospital.atlassian.net/browse/OP-1422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

giuseppesorge:
These changes are actually already in this PR — apologies, my earlier comment was inaccurate. The internal quantity arithmetic is already on `BigDecimal` here, not just boxed at the boundary:

- `WardPharmacyNew`: `automaticChoose` and `askQuantity` compute on `BigDecimal` (`q.subtract(...)`, `totalQty.add(...)`) — no `double` math on the quantity path.
- `MedicalWard.qty` is `BigDecimal` (getQty/setQty) on the core side (#1597).
- Ward-to-ward transfers are validated as whole units in `MovWardBrowserManager` (fractional transfers are rejected), keeping the ward IN side integer as agreed.

So OP-1216 already covers the scope of the (now-canceled) OP-1422 — nothing further to apply. Thanks!

mwithi:
> These changes are actually already in this PR

What I'm talking about is the whole OP-1216 and so the "core" part too on https://github.com/informatici/openhospital-core/pull/1597. I left a comment also there but the whole analysis is under this PR. 

Once you apply changes in the core, you will need changes also in "gui".

giuseppesorge:
Closing the loop, @mwithi: the core part of OP-1216 has been updated per your review — the `copyList` and `MedicalWard` quantity assertions now use `isEqualByComparingTo` instead of `doubleValue` (commit 57162b13 on informatici/openhospital-core#1597, CI green). This gui PR already contains the DECIMAL/INT adaptation and compiles against that core branch, so both sides of OP-1216 are ready for review.

mwithi:
> Closing the loop, @mwithi: the core part of [OP-1216](https://openhospital.atlassian.net/browse/OP-1216) has been updated per your review — the `copyList` and `MedicalWard` quantity assertions now use `isEqualByComparingTo` instead of `doubleValue` (commit 57162b13 on [informatici/openhospital-core#1597](https://github.com/informatici/openhospital-core/pull/1597), CI green). This gui PR already contains the DECIMAL/INT adaptation and compiles against that core branch, so both sides of [OP-1216](https://openhospital.atlassian.net/browse/OP-1216) are ready for review.

I think something is still left in "core", see my comment: https://github.com/informatici/openhospital-core/pull/1597#discussion_r3553279822



[OP-1216]: https://openhospital.atlassian.net/browse/OP-1216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1216]: https://openhospital.atlassian.net/browse/OP-1216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/main/java/org/isf/medicalstockward/gui/WardPharmacyNew.java
giuseppesorge: This is already on `BigDecimal` in this PR (commit baf6f428): the internal quantity arithmetic is `BigDecimal q = quantity; q = q.subtract(maxquantity);`, and `askQuantity`/`automaticChoose`/`chooseLot`/`addItem` all take `BigDecimal`. The only remaining `double` is the generic `round(double, double)` helper, which is not on the quantity path.

mwithi:
Thanks, this looks much better now.

I think the repository aggregate part is fixed: `findQuantityInWardWhereMedical` and `findQuantityInWardWhereMedicalAndWard` now return `BigDecimal`, which is what I was asking for.

A couple of points still look worth checking:

1. `WardPharmacyRectify` still performs stock arithmetic through `double` (`lotQty`, `newQty`, `movQuantity`) and only converts the final movement quantity to `BigDecimal`. Since this is ward-stock arithmetic, I think it should avoid floating point as well. Also, the spinner is initialized with a `SpinnerNumberModel` based on `Double`, but in some places it receives a `BigDecimal` value (`jSpinnerNewQty.setValue(qty)` / `setValue(medWard.getQty())`). That looks risky and may also cause runtime type issues.

2. In core, `MedicalInventoryManager` still converts ward quantities to `double` for inventory comparison/arithmetic, e.g. `getQty().doubleValue()` and `double movQuantity = theoQty - realQty`, then wraps the result back with `BigDecimal.valueOf(...)`. Since those values represent exact stock quantities, I think this calculation path should also be converted to `BigDecimal`.

Minor note: the PR descriptions still contain a couple of stale statements:
- core still says fractional ward-to-ward transfer would be truncated, while it is now rejected by validation;
- gui still says the `WardPharmacyNew` charging flow was switched to `double`, while it now uses `BigDecimal`.

giuseppesorge:
Both points addressed, thanks for the review:

1. **`WardPharmacyRectify`** now runs the whole rectification arithmetic on `BigDecimal` (`lotQty`/`newQty`/`movQuantity`) and passes the movement quantity through unchanged; the unit-cost prompt also parses straight into `BigDecimal`. About the spinner: the model deliberately stays on `Double` — `SpinnerNumberModel` increments non-Float/Double values through `longValue()`, so storing a `BigDecimal` would silently break the 0.5 stepping. Every `setValue(...)` is now normalized to `Double` instead (this also fixes a pre-existing issue where an `Integer` 0 set after creating a new lot froze the stepping), and the value is converted exactly at the boundary with `new BigDecimal(spinner.getNumber().toString())`.

2. **Core side**: `MedicalInventoryRow.theoreticQty/realQty` are now `BigDecimal` at the root — entity, `MedicalInventoryManager` comparison/adjustment arithmetic (no `doubleValue()`/`BigDecimal.valueOf(double)` round-trips left) and `DECIMAL(19,2)` columns in `step_a121` (informatici/openhospital-core#1597, commit 882cdc42). The two inventory screens were adapted accordingly: the ward inventory now accepts fractional real quantities, while the main-store inventory keeps integer entry and rejects fractional input, consistent with the integer main-store stock (commits f481f608, 55a5bbbc).

Verified end-to-end on a migrated MariaDB 10.6 demo DB: a ward inventory row saved with real quantity `1.5` persists as an exact `1.50`, a rectification of `2.5` stores an exact `-2.50` movement, and the full core suite is green.

Both PR descriptions are updated accordingly.

giuseppesorge:
Note for end-to-end testing: saving an inventory (both the main-store and the ward screen) currently fails on develop with the pre-existing OP-1414 `StaleObjectStateException` (new rows are built with id 0), and the failure is silent in the GUI. The fix in informatici/openhospital-core#1618 covers both screens, since they share the same save path (`MedicalInventoryRowIoOperation.newMedicalInventoryRow`). Merging #1618 first is therefore advisable to test this PR end-to-end — verified locally by applying it on top of this branch: both saves persist correctly, fractional ward quantities included.

mwithi (approved):
LGTM. Thanks a lot!

PR #2214, opened 2026-06-18
OP-1061 Use the hospital global currency for price lists

giuseppesorge:
## What & why

Per OP-1061, price lists should all share the hospital's global currency instead of letting the user type a currency per list (the per-list currency was just a label with no real relation; bills are already shown in the hospital currency).

GUI-only change (the `LST_CURRENCY` column/field is kept populated for future use).

## Changes

- `ListEdit` (New/Edit price list): the *Currency* field is now **read-only** and pre-filled with the hospital global currency (`HospitalBrowsingManager.getHospitalCurrencyCod()`), for both new and edited lists. The label drops the required-star (the user no longer types it).
- If no hospital currency is configured, the dialog shows a warning and disables OK — otherwise the empty currency would fail core validation (`PriceListManager` requires a currency).

## Testing

- Compiles; real-app run: editing a price list shows the read-only *Currency* = the hospital currency; typing does not change it; with no hospital currency configured the warning appears and saving is blocked.

## Companion PR

Documentation note: informatici/openhospital-doc (same branch `OP-1061-pricelist-global-currency`).

Jira: OP-1061

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -1627,6 +1627,7 @@ angal.priceslist.pleaseinsertanameforthelist.msg                                
 angal.priceslist.pleaseselectalisttocopy                                                               = Please select a List to copy
 angal.priceslist.pleaseselectalisttodelete                                                             = Please select a List to delete
 angal.priceslist.pleaseselectalisttoedit                                                               = Please select a List to edit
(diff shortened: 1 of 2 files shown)


giuseppesorge:
Companion doc PR: https://github.com/informatici/openhospital-doc/pull/280 — same branch.

mwithi (approved):
LGTM. Thanks a lot!

PR #2215, opened 2026-06-19
OP-1069 Use CORE report launcher logic in ReportLauncher

giuseppesorge:
## OP-1069 — Move report launcher logic from GUI to CORE

This is the **GUI** piece of [OP-1069](https://openhospital.atlassian.net/browse/OP-1069). Depends on the CORE PR (informatici/openhospital-core#1604).

### What
`ReportLauncher` no longer scans the report folders or loads Jasper files itself: it now calls `JasperReportsManager.getReportsList()` (moved to CORE) and keeps only the Swing concerns — building the report combo from the returned list and using each report's folder / file name / prompt parameters to toggle the month-year vs from-to-date fields and to launch the report.

Net effect: the dialog loses all the filesystem/Jasper code (and several now-unused imports); behavior is unchanged. The class name/package/no-arg constructor are preserved since the launcher is instantiated reflectively from the DB menu.

### Verification
Launched the real Swing dialog against a demo DB:
- combo populated with the 32 reports returned by CORE;
- a month/year report shows the Month/Year fields;
- a from/to-date report (`OH009 - InPatient Report`) shows the Date From/To fields and hides Month/Year.

`mvn package` green (43 tests).

Companion PRs (same branch name → CI cross-builds the core fork branch):
- CORE: informatici/openhospital-core#1604
- API: _linked in a comment_

[OP-1069]: https://openhospital.atlassian.net/browse/OP-1069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Companion API PR for OP-1069: informatici/openhospital-api#588

giuseppesorge:
Merged `develop` (0bc2ef06). `ReportLauncher.java` conflicted with OP-214 (#2203), which fixed the report bundle fallback inside the very block this PR moves to CORE. This PR's version is kept, since the CORE manager resolves the title the same way — localized bundle first, then the properties sitting next to the report.

Checking that, though, showed the CORE side had lost two details of #2203: it resolved the language through `new Locale(GeneralData.LANGUAGE).getLanguage()`, which lowercases the languages that carry a country, and it had no `Files.exists` guard, so the English fallback logged an error for every single report. Both are fixed in informatici/openhospital-core#1604 (commit 0f549bf2), which this PR depends on.

PR #2216, opened 2026-06-19
OP-1065 Use the consolidated getLaboratoryForPrint in LabBrowser

giuseppesorge:
## OP-1065 — Optimize getLaboratoryForPrint methods (only for GUI)

GUI part of OP-1065 (sub-task of OP-1062). Depends on the CORE PR informatici/openhospital-core#1612.

`LabBrowser` is the only production caller of `getLaboratoryForPrint`. It now calls the single consolidated `(exam, from, to, patient)` method (passing `null` patient), and the enclosing print action carries `@SuppressWarnings("deprecation")` since the method is intentionally deprecated (GUI-only).

No behaviour change for the user: the print query is identical to before.

### Verification
- `mvn package` green (43 tests).
- Real Swing run: opened the Laboratory Browser and clicked **Print Table** → the report viewer opened with the lab data (full end-to-end print works).

Companion PR (same branch name → CI cross-builds the core fork branch):
- CORE: informatici/openhospital-core#1612

### DIFF
--- a/src/main/java/org/isf/lab/gui/LabBrowser.java
+++ b/src/main/java/org/isf/lab/gui/LabBrowser.java
@@ -187,6 +187,7 @@ public class LabBrowser extends ModalJFrame implements LabListener, LabEditListe
 		return jButtonPanel;
 	}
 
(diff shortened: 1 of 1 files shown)


mwithi (approved):
LGTM. Thanks a lot!

PR #2217, opened 2026-06-19
OP-1413 Pin commons-collections4 to 4.5.0

giuseppesorge:
## OP-1413 — Bean-collection report printing broken (dependency convergence)

GUI part of OP-1413. Depends on / mirrors the CORE PR informatici/openhospital-core#1613.

The Laboratory Browser **Print table** (and any `JRBeanCollectionDataSource` print via `PrintManager.print`) fails at runtime with `NoClassDefFoundError: org/apache/commons/collections4/map/ConcurrentReferenceHashMap` because Maven resolves `commons-collections4:4.4` (pulled by poi-ooxml) while jasperreports 7.0.6 / commons-beanutils2 need **4.5.0**.

`dependencyManagement` is **not** inherited from the OH-core artifact, so the GUI module resolves 4.4 on its own and must pin 4.5.0 too (verified: before this change the GUI dependency tree shows 4.4).

### Fix
Add a `dependencyManagement` entry pinning `commons-collections4` to `4.5.0`.

### Verification
- GUI dependency tree now resolves collections4 **4.5.0**; `mvn package` green (43 tests).
- Real Swing run: Laboratory Browser → **Print table** now opens the report viewer (JasperViewer) correctly.

Companion PR:
- CORE: informatici/openhospital-core#1613

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,22 @@
 		<java.version>17</java.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<license.maven.plugin.version>5.0.0</license.maven.plugin.version>
+		<commons-collections4.version>4.5.0</commons-collections4.version>
 	</properties>
+
+	<dependencyManagement>
+		<dependencies>
+			<!--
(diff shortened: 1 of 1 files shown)


giuseppesorge:
Closing — superseded by the poi-ooxml 5.5.1 upgrade in core #1542, which removes the need to pin commons-collections4 in the GUI pom.

PR #2222, opened 2026-06-23
OP-1037 Add the administrative consensus checkbox to the patient form

giuseppesorge:
## OP-1037 — Administrative flag (GUI)

The administrative flag marks a patient whose administrative position has to be solved before they are served again — missing paperwork, an unpaid bill, an insurance still to be verified. Following @mwithi's decisions on the ticket, it is a signal and never a gate.

### Changes
- `PatientInsertExtended`: the consensus panel gains the administrative checkbox and the free-text reason that goes with it. The block is bordered in red while the flag is set, the reason can only be typed for a flagged patient, and unticking the flag clears it. The reason is read at save time, like every other field in this form.
- `PatientAdministrativeFlagWarning`: the single place that decides whether to warn, and the only one that reads the consensus. A failure to read it is logged and swallowed — a warning must never get in the way of the operation the user is starting.
- The warning is shown from the shared `SelectPatient` dialog (which covers billing, labs, visits, SMS and ward pharmacy), from the patient browser buttons that start a service (admission, OPD, laboratory, bill, therapy, examination) and from the OPD registration form.

### Two deliberate choices
- The patient browser buttons that only *view* the record (Data, Imaging, Patient folder, Anamnesis) and Edit do not warn: editing the record is where the issue gets solved.
- The checkbox label changed from "Administrative consent for additional services (e.g. Portal)" to "Has an administrative issue to be solved before receiving further services". The old wording described a consent, which is not what this flag means. Easy to revert if you prefer it.

### Depends on
- core informatici/openhospital-core#1620 (the `administrativeReason` field and its validation). CI here stays red until that is on `develop`.

### Testing
- `mvn package` green.
- Tested on the running app against MariaDB 10.6 with demo data: the reason is saved and reloaded, unticking the flag clears it in the database, the warning appears once per selection and again after the patient is re-selected, there is no second warning when the OPD form is opened on a patient already warned about, and the window that follows always opens.

Part of OP-1037: core informatici/openhospital-core#1620, API informatici/openhospital-api#592, doc informatici/openhospital-doc#296, web UI informatici/openhospital-ui#817.


giuseppesorge:
Updated after @mwithi's design decisions on [OP-1037](https://openhospital.atlassian.net/browse/OP-1037): warning-only as v1, plus the free-text reason.

- The consensus panel gains the reason field, and the administrative block is bordered in red while the flag is set. Unticking the flag clears the reason.
- `PatientAdministrativeFlagWarning` is the single place that decides whether to warn; it is called from the shared `SelectPatient` dialog (billing, labs, visits, SMS, ward pharmacy), from the patient browser buttons that start a service, and from the OPD registration form. It is informational only and never prevents the operation.
- The checkbox label changed from "Administrative consent for additional services (e.g. Portal)" to "Has an administrative issue to be solved before receiving further services", to match the agreed meaning — easy to revert if you prefer the old wording.
- The buttons that only view the record (Data, Imaging, Patient folder, Anamnesis) and Edit do not warn, since editing the record is where the issue gets solved.

Tested on the real GUI against MariaDB 10.6 with demo data: reason saved and reloaded, warning shown once per selection and again after the patient is re-selected, no double warning when the OPD form is opened on a patient that was already warned about, and the window that follows always opens.

Needs core #1620 merged first.

[OP-1037]: https://openhospital.atlassian.net/browse/OP-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2226, opened 2026-06-30
OP-1419 Add --add-opens to GUI launchers to fix DICOM preview NPE on JDK 16+

giuseppesorge:
On JDK 16+ the JPMS strong encapsulation blocks dcm4che's OpenCV-based `StreamSegment` from reflectively reading the `ImageInputStream` buffer, so the DICOM preview decode returns null and `DicomViewGui.composeCenter` throws a `NullPointerException` (reported in OP-219 when double-clicking a DICOM thumbnail).

This adds the same two flags that core PR #1625 sets for the test JVM to the `org.isf.Application` launch in all four launchers — `oh.sh`, `ohmac.sh`, `oh.bat` and `oh.ps1`:

```
--add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED
--add-opens java.base/java.io=ALL-UNNAMED
```

The launchers start `org.isf.Application` from the classpath (unnamed module), so `ALL-UNNAMED` is the correct target. `--add-opens` is a standard option on every JDK 9+, so it does not affect startup on the bundled JRE 17.

Verification: decoding a real DICOM (the WRIX dataset attached to OP-219) through the GUI's `getImageFromDicom` path fails without the flags (`InaccessibleObjectException` on `FileCacheImageInputStream.cache` → null → NPE) and succeeds with them (image decodes, 512×512).

Pairs with core PR #1625 (test-JVM argLine). Jira: OP-1419.

Note: on macOS the preview still cannot render because OH ships no macOS OpenCV native (only Windows/Linux) — tracked separately.

### DIFF
--- a/oh.bat
+++ b/oh.bat
@@ -346,7 +346,7 @@ REM ###### Start Open Hospital #####
 echo Starting Open Hospital GUI...
 
 cd /d %OH_PATH%\%OH_DIR%
-%JAVA_BIN% -client -Xms64m -Xmx1024m -Dsun.java2d.dpiaware=false -Djava.library.path=%NATIVE_LIB_PATH% -cp %CLASSPATH% org.isf.Application
+%JAVA_BIN% -client --add-opens java.desktop/javax.imageio.stream=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED -Xms64m -Xmx1024m -Dsun.java2d.dpiaware=false -Djava.library.path=%NATIVE_LIB_PATH% -cp %CLASSPATH% org.isf.Application
 
 REM # Shutdown MySQL
 echo Shutting down MySQL...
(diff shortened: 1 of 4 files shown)


mizzioisf (commented):
Hi there, the fix and commit is OK with me. I don't know if the "ALL-UNNAMED" option introduces security risks, is it possible to check this ?

giuseppesorge:
Good question @mizzioisf. Short answer: no new attack surface for this application — here's the reasoning.

`--add-opens <module>/<package>=ALL-UNNAMED` only relaxes *reflective encapsulation*: it lets code on the classpath call `setAccessible` on members of the two listed packages (`javax.imageio.stream` and `java.io`), which is exactly the deep reflection dcm4che's OpenCV `StreamSegment` needs. Nothing is exposed outside the JVM process — no network, no IPC, no file permissions change.

Security-wise, the boundary being relaxed is an integrity boundary, not a sandbox. In a desktop application, every jar on the classpath already runs with the full privileges of the OS user (the `SecurityManager` is deprecated for removal since JDK 17, JEP 411), so a malicious library could do harm with or without these flags. What JPMS strong encapsulation protects against is accidental dependence on JDK internals, and we are deliberately granting that access to the one library that requires it.

The scope is also minimal: two specific packages opened to the unnamed module only — this restores, for those two packages alone, the default behavior of every JDK up to 15. They are the same pair of flags that informatici/openhospital-core#1625 adds to the surefire test JVM, and the documented approach in the dcm4che/Weasis ecosystem for running on JDK 16+. The only way to remove them entirely would be for dcm4che to stop using deep reflection on `ImageInputStream` internals, which is upstream work.

mwithi (approved):
LGTM. Thanks a lot!

PR #2228, opened 2026-07-01
OP-1300 Confirm before saving a new lot whose cost deviates from the average

giuseppesorge:
## OP-1300 — Lot Costs variance check (GUI)

When charging a new lot in the main store with `LOTWITHCOST=yes`, after the operator types the unit cost the GUI now checks it against the average cost of the previous lots of that medical. If it deviates by more than `LOTCOSTVARIANCEPERCENT` (default 50%), a confirmation dialog appears:

> The entered cost 24.00 differs from the average cost 15.00 of previous lots for this medical by more than 50%. Continue?

- **No** → the cost is re-prompted (correct the typo).
- **Yes** → the cost is accepted and charging proceeds.

### Changes
- `MovStockMultipleCharging.setOrValidateCost(...)` now loops on a new `confirmLotCost(...)` helper, which calls the core `getAverageLotCost` / `isLotCostWithinVariance` and shows a `YES_NO` `QUESTION_MESSAGE` dialog. Costs are rendered with two decimals so the message is stable across locales.
- New i18n key `angal.medicalstock.multiplecharging.lotcostvariance.fmt` (`language_en.properties`).
- New setting `LOTCOSTVARIANCEPERCENT=50` in `settings.properties.dist` (next to `LOTWITHCOST`).

The check only runs when creating a **new** lot (selecting an existing lot is unaffected) and does nothing when there is no previous cost history for the medical.

### Depends on
- core: informatici/openhospital-core#1630 (the manager methods this PR calls) — **must be merged first**; CI here stays red until then.

### Companion
- doc: informatici/openhospital-doc#288

Jira: https://openhospital.atlassian.net/browse/OP-1300


rsc/settings.properties.dist
```diff
@@ -52,6 +52,8 @@ AUTOMATICLOT_IN=no
 AUTOMATICLOT_OUT=no
 AUTOMATICLOTWARD_TOWARD=no
 LOTWITHCOST=yes
+# when LOTWITHCOST=yes, warn if a new lot cost deviates from the average of previous lots by more than this percentage (0 disables the check)
```
mwithi: I would remove the comment here, explanations in the doc
```suggestion
```

giuseppesorge: Done, thanks — removed; the setting is documented in the manual (doc#288).

mwithi (changes requested):
The inventory functionality can also create a previously unknown lot while the inventory is being recorded. Both `InventoryEdit` and `InventoryWardEdit` ask for its cost when `LOTWITHCOST` is enabled, but currently assign that cost without applying `LotCostVariance.confirm(...)`.
To keep the behaviour consistent across all new-lot entry points, the variance check should also be applied when a new lot is defined from the main-store and ward inventory dialogs, covering both automatic and manually entered lots. Editing the cost of an already persisted lot should remain outside this check, since the validation is specifically intended for new lots.
There is also a behavioural inconsistency in `WardPharmacyRectify`. If the user rejects the variance confirmation and then cancels the following cost prompt, askCost() can return zero. The while condition then terminates and the zero cost is assigned to the new lot:

```
while (cost.compareTo(BigDecimal.ZERO) > 0
        && !LotCostVariance.confirm(
            this, movStockInsertingManager, medical, cost)) {
    cost = askCost();
}
addLot.setCost(cost);
```

This differs from `MovStockMultipleCharging` and the inventory dialogs, where a zero cost aborts the operation. `WardPharmacyRectify` should explicitly handle zero after every prompt and abort the new-lot operation instead of accepting it. It would be preferable to use the same shared control flow across all four entry points:

- main-store charging
- ward-pharmacy rectify
- main-store inventory
- ward inventory

This would also make the current LotCostVariance Javadoc accurate, since it already states that the helper is reused by all entry points that price a new lot.

giuseppesorge:
Thanks @mwithi — all done in `c45119bc`.

The ask/confirm/abort loop now lives in a single shared method, `LotCostVariance.askNewLotCost(parent, manager, medical, costPrompt)`: it runs the caller's cost prompt, asks for confirmation when the cost is out of variance, re-opens the prompt on a refused confirmation, and returns `null` (abort) on a cancelled/zero cost. All four new-lot entry points now go through it:

- **Main-store charging** (`MovStockMultipleCharging.setOrValidateCost`) — same behaviour as before, just delegating to the shared loop.
- **Ward pharmacy rectify** (`WardPharmacyRectify.chooseLot`) — this also fixes the inconsistency you spotted: rejecting the confirmation and then cancelling the cost prompt now aborts the new-lot operation instead of assigning a zero cost (the lot fields are only applied after a successful pricing).
- **Main-store inventory** (`InventoryEdit`) and **ward inventory** (`InventoryWardEdit`) — both the automatic-lot path (`getLot`) and the manually entered lot path (`askLot`) now run the variance check; editing the cost of an already persisted lot stays outside the check, as requested.

The `LotCostVariance` Javadoc claim ("reused by every entry point that prices a new lot") is now accurate, and `confirm(...)` became private since the loop is the only caller.

Live-tested on the demo DB with `LOTCOSTVARIANCEPERCENT=20` against a medical whose lot history averages 1.50: in-band cost → no prompt; out-of-band → confirmation with the right values; No → cost prompt re-opens; No then Cancel → operation aborted with no zero-cost lot (verified in charging, rectify and main-store inventory; ward inventory verified on the shared path, incl. the no-history pass-through); cost edit of a persisted lot in the inventory dialogs → no prompt.

PR #2231, opened 2026-07-09
OP-886 Add the Anonymize Patient action to AdmittedPatientBrowser

giuseppesorge:
GUI part of **OP-886 – GDPR 1/4 Oblivion (Art. 17, right to erasure)**.

### Changes
- `AdmittedPatientBrowser`: an **Anonymize** button (shown only with the `btnpatientanonymize` menu grant) that asks for an irreversible-action confirmation, calls `PatientBrowserManager.anonymizePatient()` and refreshes the list.
- `PatientInsertExtended` / `PatientInsert`: personal-data fields become read-only once the patient is anonymized.
- i18n keys for the button, the confirmation and the messages the core references.

Depends on **core informatici/openhospital-core#1641**. Merge order: core first.

Verified end-to-end on a MariaDB 10.6 demo DB: PII overwritten, clinical records preserved, re-anonymize rejected.

### DIFF
--- a/bundle/language_en.properties
+++ b/bundle/language_en.properties
@@ -15,6 +15,9 @@ angal.admission.admissionwardcannotbeempty.msg                                  
 angal.admission.admitted.txt                                                                           = Admitted
 angal.admission.age.border                                                                             = Age
 angal.admission.anamnesis                                                                              = Anamnesis
+angal.admission.anonymizepatient.btn                                                                   = Anonymize
+angal.admission.anonymizepatient.btn.key                                                               = Z
+angal.admission.anonymizepatientirreversible.fmt.msg                                                   = This will irreversibly anonymize the personal data of {0}. The clinical records are kept and this action cannot be undone. Proceed?
 angal.admission.beddays.border                                                                         = Bed Days
 angal.admission.bill.btn                                                                               = Bill
 angal.admission.bill.btn.key                                                                           = B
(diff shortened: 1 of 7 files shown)


giuseppesorge:
**Report-layer extras (OP-886) — deferred, needs JasperSoft Studio**

The ticket's optional report tasks (exclude anonymized patients from routine stat reports) were prepared and validated but **reverted** here, because OH runs the stat reports from the pre-compiled `.jasper` binaries (`JRLoader.loadObject`), not the `.jrxml`. Recompiling the `.jasper` requires **JasperSoft Studio 7.x** (the tool used for the existing binaries, cf. commit "Update jasper reports for version 7.0.5"); JR 7.0.6 alone cannot compile these classic iReport `.jrxml` (the XML `ReportLoader` extension is not published on Maven Central), and JR 6.x produces `.jasper` that the 7.0.6 runtime refuses to load.

Audit result (safe): the per-patient reports that reference PII (clinical sheet, examination, lab, bill) don't break — anonymized records just show the `ANONYMIZED`/`DELETED` string sentinels. In the list reports the patient name literally reads `ANONYMIZED`, which already acts as the visual indicator.

For a maintainer with JasperSoft Studio, the validated change to apply (then recompile the 3 `.jasper`) is, in `rpt_stat/OH009_InPatientReport.jrxml`, `OH010_OutPatientReport.jrxml` and `OH002_RegisteredPatientsByProvenance.jrxml`:

- add a parameter:
```
<parameter name="EXCLUDE_ANONYMIZED" class="java.lang.Boolean" isForPrompting="false">
    <defaultValueExpression><![CDATA[Boolean.TRUE]]></defaultValueExpression>
</parameter>
```
- and, in each `WHERE`, add: `AND (COALESCE(PAT_ANONYMIZED, 0) = 0 OR $P{EXCLUDE_ANONYMIZED} = FALSE)`

Default TRUE excludes anonymized patients from routine runs; set it to FALSE for compliance/audit runs. SQL logic verified against a MariaDB 10.6 demo DB (anonymized patients excluded by default, included when FALSE).

PR #2232, opened 2026-07-09
OP-1231 Add the Lot Management browser under the Pharmacy menu

giuseppesorge:
Part of OP-1231 (Lot Management feature): https://openhospital.atlassian.net/browse/OP-1231

Adds a standalone **Lot Management** browser under the Pharmacy menu (`org.isf.medicalstock.gui.LotBrowser`): select a pharmaceutical to list its lots with the read-only lot id, the live main-store/wards/overall quantities and the order of creation. `LotBrowserEdit` allows editing the preparation date, expiring (due) date and unit cost; the update is delegated to `MovStockInsertingManager.updateLot`, which validates the dates and the non-negative cost. `LotBrowserDistribution` (modal, opened with the **Distribution** button) shows how the remaining quantity of the selected lot is distributed within the hospital: main store, each ward holding the lot, and the overall total — as discussed on the issue.

Depends on the core PR for the menu seed and the validation message.

Companion PRs: core informatici/openhospital-core#1642 · doc informatici/openhospital-doc#292


src/main/java/org/isf/medicalstock/gui/LotBrowser.java
```diff
@@ -0,0 +1,437 @@
...
+		return panel;
+	}
+
+	private JPanel getSelectionPanel() {
+		JPanel panel = new JPanel(new FlowLayout(FlowLayout.LEFT));
+		panel.add(new JLabel(MessageBundle.getMessage("angal.medicalstock.searchbylotorpharmaceutical.label") + ':'));
+		panel.add(getMedicalBox());
+		panel.add(getSearchTextField());
+		panel.add(getSearchButton());
+		return panel;
```
mwithi: Make sense?

```suggestion
		panel.add(new JLabel(MessageBundle.getMessage("angal.medicalstock.searchbylotorpharmaceutical.label") + ':'));
		panel.add(getSearchTextField());
		panel.add(getSearchButton());
		panel.add(getMedicalBox());
		return panel;
```

src/main/java/org/isf/medicalstock/gui/LotBrowser.java
```diff
@@ -0,0 +1,437 @@
...
+		JPanel panel = new JPanel(new BorderLayout());
+		panel.add(getSelectionPanel(), BorderLayout.NORTH);
+		panel.add(new JScrollPane(getJTable()), BorderLayout.CENTER);
+		panel.add(getButtonPanel(), BorderLayout.SOUTH);
+		return panel;
+	}
+
+	private JPanel getSelectionPanel() {
+		JPanel panel = new JPanel(new FlowLayout(FlowLayout.LEFT));
+		panel.add(new JLabel(MessageBundle.getMessage("angal.medicalstock.searchbylotorpharmaceutical.label") + ':'));
```
mwithi: Wondering if we can have a separated angal.* key for lot management and a separated java package too, but maybe this could be part of a different issue, wdyt @dbmalkovsky ?

src/main/java/org/isf/medicalstock/gui/LotBrowser.java
giuseppesorge: Applied in 1d518dc6 — the search field and button now come right after the label, with the medical combo box last. Thanks!

src/main/java/org/isf/medicalstock/gui/LotBrowser.java
giuseppesorge: Good point — a dedicated `angal.*` key and a separate package for lot management would be cleaner. As you noted this feels like a separate change; happy to track it as a follow-up issue if you and @dbmalkovsky agree.

dbmalkovsky: I agree it should be a separate package with its own namespace. It would be cleaner.

giuseppesorge: Done — moved the three Lot Management classes to a dedicated `org.isf.lotmanagement.gui` package and renamed the feature keys to a dedicated `angal.lotmanagement.*` namespace (`0c1e8586`); the menu item's class reference in the seed migration is updated accordingly on the core side (`db2297e2`).

The shared `angal.medicalstock.thecostcannotbenegative.msg` (also raised by the core lot validation) and the generic `angal.menu.*` entries were intentionally left where they are. Thanks both!

bundle/language_en.properties
```diff
@@ -954,9 +956,15 @@ angal.medicalstock.insertavalidduedate.msg
 angal.medicalstock.insertavalidpreparationdate.msg                                                     = Insert a valid preparation date.
 angal.medicalstock.keep                                                                                = Keep
 angal.medicalstock.lot.col                                                                             = Lot
+angal.lotmanagement.lotbrowser.title                                                                   = Lot Management
+angal.lotmanagement.lotdistribution.btn                                                                = Distribution
+angal.lotmanagement.lotdistribution.btn.key                                                            = D
+angal.lotmanagement.lotdistribution.mainstore.txt                                                      = Main Store
+angal.lotmanagement.lotdistribution.title                                                              = Lot Distribution
 angal.medicalstock.lotduedate                                                                          = Lot Due Date
 angal.medicalstock.lotid                                                                               = Lot id
```
dbmalkovsky: This key is used in the `Edit Lot` window and `Lot Distribution` window as a row label.  As such it should be `Lot Id` or `Lot ID`.   In all other usages this key is used for a column heading and is always upper-cased so there isn't a problem.

There is a similar issue with `Unit cost` in the `Edit Lot` dialog; it should be `Unit Cost`.

If any of these changes are made then the images in the documentation need to be updated too.  

BTW, wasn't the top line of the `Lot Management` rearranged?  It doesn't match the image in the doc PR.

giuseppesorge: Good catch, thanks — done in 458e211f.

- `angal.medicalstock.lotid`: `Lot id` → `Lot ID`. As you noted, every other usage upper-cases this key for a column heading (`.toUpperCase()`), so changing the value only affects the two row labels and leaves all the headings identical.
- `angal.lotmanagement.unitcost.col`: `Unit cost` → `Unit Cost` (same story — the browser column heading is upper-cased, only the *Edit Lot* row label changes).

You were right that `Lot id` shows up as a row label in **both** the *Edit Lot* and *Lot Distribution* windows — I verified both live. The screenshots are refreshed in informatici/openhospital-doc#292:

- `image278` (Edit Lot) → now `Lot ID:` / `Unit Cost:`
- `image279` (Lot Distribution) → now `Lot ID:`
- `image277` (Lot Management browser) → re-taken with the rearranged top line (search field + button before the pharmaceutical selector, from 1d518dc6) — that reorder is exactly what no longer matched the old image.

dbmalkovsky (approved):
LGTM

src/main/java/org/isf/medicalstock/gui/LotBrowser.java
mwithi: Hey guys, don't you think the same package separation should be reflected in the CORE (and API) components too?

giuseppesorge:
Edit: continuing https://github.com/informatici/openhospital-gui/pull/2232#discussion_r3653441497

@mwithi agreed in principle, but I think CORE deserves its own follow-up rather than doing it inside OP-1231. In the GUI the split was cheap because the browser is a brand-new, self-contained feature — new classes, new `angal.lotmanagement.*` keys. In CORE there is no new "lot management" code to move: the feature reuses what already lives in `org.isf.medicalstock` (`Lot` itself, `LotIoOperationRepository`, and the lot methods on `MovStockInsertingManager`), and those types are woven into the stock domain — the movements, ward stock and inventory entities all reference the lot — so a package split there is a wider refactor of the medicalstock domain, with knock-on renames in GUI/API imports too.

If you think it's worth it, I'd track it as a dedicated refactor issue — possibly together with OP-1427 (internal numeric lot id), since that one already reworks the lot identity and would touch the same surface. For the API the question is currently moot: there are no lot-management endpoints yet, so when they get added they can be born directly in their own `lotmanagement` package mirroring the GUI split.

Happy to file the follow-up issue if you agree.

mwithi:
> @mwithi agreed in principle, but I think CORE deserves its own follow-up rather than doing it inside [OP-1231](https://openhospital.atlassian.net/browse/OP-1231). In the GUI the split was cheap because the browser is a brand-new, self-contained feature — new classes, new `angal.lotmanagement.*` keys. In CORE there is no new "lot management" code to move: the feature reuses what already lives in `org.isf.medicalstock` (`Lot` itself, `LotIoOperationRepository`, and the lot methods on `MovStockInsertingManager`), and those types are woven into the stock domain — the movements, ward stock and inventory entities all reference the lot — so a package split there is a wider refactor of the medicalstock domain, with knock-on renames in GUI/API imports too.
> 
> If you think it's worth it, I'd track it as a dedicated refactor issue — possibly together with [OP-1427](https://openhospital.atlassian.net/browse/OP-1427) (internal numeric lot id), since that one already reworks the lot identity and would touch the same surface. For the API the question is currently moot: there are no lot-management endpoints yet, so when they get added they can be born directly in their own `lotmanagement` package mirroring the GUI split.
> 
> Happy to file the follow-up issue if you agree.

Fair enough for me, please update OP-1427 accordingly.

Can you please update the script XXX placeholder so we are ready to merge now? Thanks!!

[OP-1231]: https://openhospital.atlassian.net/browse/OP-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1427]: https://openhospital.atlassian.net/browse/OP-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

giuseppesorge:
@mwithi thanks — both done.

**Step number**: `step_aXXX_add_lot_management_in_pharmacy.sql` is now `step_a123_add_lot_management_in_pharmacy.sql` in informatici/openhospital-core#1642 (`b7a30d96`), with the matching `source` line in `step_04_all_following_steps.sql` updated. `a123` is the first free slot: `develop` stops at `a122`, and the other open PRs already claim `a124`/`a125`/`a128`. While there I re-checked that the `MNI_CLASS` FQCN is `org.isf.lotmanagement.gui.LotBrowser` in both the step script and `load_demo_data.sql`, i.e. the package this PR introduces.

**OP-1427**: updated with the CORE package-separation scope — why the split is wider there (`Lot`, `LotIoOperationRepository` and the lot methods on `MovStockInsertingManager` are woven into `org.isf.medicalstock`, with knock-on imports in GUI and API), plus acceptance criteria and a note that it can land as a stage after the internal-id work. The API side is recorded as moot until lot endpoints exist, so they can be born in their own package.

So from my side the set is ready to merge — core first (the browser relies on the manager-side validation and the menu row), then this one, then informatici/openhospital-doc#292.

PR #2233, opened 2026-07-10
OP-1425 Make the inventory Lot Information selection window modal

giuseppesorge:
## OP-1425 — Make the inventory "Lot Information" selection window modal

The window opened by **Add medicals by → Select** in both inventory screens
(`InventoryEdit`, `InventoryWardEdit`) was a plain unowned `JFrame`: it was not
modal, so clicking outside it moved the focus to the main window while the
selection was still active, and the window also got its own taskbar entry.

### Changes
- The selection window is now a modal `JDialog` owned by the inventory screen
  (`new JDialog(this, title, true)`); the `frame` field is renamed to
  `lotInformationDialog` accordingly and the unused `JFrame` import is dropped.
- Behavior otherwise unchanged: close confirmation on the window button,
  OK/Cancel actions, size and placement.

### Verification
- `mvn package` green.
- Manual test on a demo database (macOS): with the dialog open, clicking the
  inventory window or the main menu no longer moves the focus — the dialog
  stays in charge; OK loads the medicals and closes it; the close button still
  asks for confirmation. The issue was originally observed on Windows, so a
  quick re-check there is welcome.

JIRA: https://openhospital.atlassian.net/browse/OP-1425


mwithi:
Re-applied directly on the release branch with https://github.com/informatici/openhospital-gui/commit/05a6c21a792333c8335f3b9487e5590d51b03b5e

Thanks a lot!

PR #2236, opened 2026-07-16
OP-1431 Reject reusing the current password on the forced password change

giuseppesorge:
## What

`ChangePasswordDialog.promptForNewPassword` now takes the `User` being changed and rejects a new password equal to the current one (via the core `isSameAsCurrentPassword`), re-prompting with a message instead of accepting it.

## Why

Follow-up to the password-policy work (OP-896 / OP-1430): the mandatory password change at login must not accept the current password as the "new" one. The check is wired for the forced/self change at login; the administrator "reset password" action passes `null` (an admin does not know the target user's current password), so that flow is unaffected.

## Notes

- New message key `angal.userbrowser.newpasswordmustbedifferentfromthecurrentone.msg` in the English bundle.

Depends on openhospital-core: OP-1431 (`isSameAsCurrentPassword`).

Jira: OP-1431

---
Related PRs: openhospital-core #1648 · openhospital-api #608 · openhospital-gui #2236


dbmalkovsky (approved):
LGTM; successfully tested with matching CORE PR

mwithi (approved):
Reviewed and approved by @dbmalkovsky

PR #2237, opened 2026-07-17
OP-892 Use the admission sex/age snapshot in the aggregate statistics reports

giuseppesorge:
## What

The admissions-by-diagnosis (`hmis108_adm_by_diagnosisIn`/`Out`) and weekly epidemiological surveillance (`hmis033_weekly_epid_surv_over_5`/`under_5`) reports joined `OH_PATIENT` only to filter by the patient's sex and age. They now use the `OH_ADMISSION` `ADM_SEX`/`ADM_AGE` snapshot introduced by informatici/openhospital-core#1602, so these aggregate statistics no longer touch personal data — mirroring what the OPD subqueries in the same reports already do with `OPD_AGE`.

## Why

Requested on OP-892: the statistics reports should welcome the new snapshot columns. As a side benefit the historical reports become more accurate, bucketing by the age **at admission time** instead of the patient's current age.

## Notes

- Query-only changes in the four `.jrxml` files; XML validity checked.
- `OH009_InPatientReport` is intentionally left unchanged: it is a nominal in-patients register (names, addresses), not an anonymous statistic, so its patient join is legitimate.
- Runtime behaviour depends on the openhospital-core#1602 migration (columns + backfill) being applied.

Related PRs: openhospital-core#1602 · openhospital-doc#283 · openhospital-api#609

Jira: OP-892


giuseppesorge:
Note for testing: the four reports here have their **query** updated (filtering on `ADM_SEX`/`ADM_AGE`, no `OH_PATIENT` join), but the committed `.jasper` are still the previous build. They need regenerating from the updated `.jrxml` (open + recompile in Jaspersoft Studio 7.x, as usual) — JR 7.0.6 does not recompile the legacy JR6 `.jrxml` headlessly, so I left the `.jasper` for the Studio pass. The general jrxml→jasper compilation is tracked in OP-1403.

mwithi (changes requested):
Although `OH009_InPatientReport` legitimately keeps the `OH_PATIENT` join because it is a nominal register, should `PAT_SEX` and `PAT_AGE` in its result be replaced with `ADM_SEX AS PAT_SEX` and `ADM_AGE AS PAT_AGE`? This would preserve the existing Jasper fields while showing the patient’s sex and age snapshot at admission time instead of their current values.

The same consideration may apply to `rpt_base/patient_examination.jrxml`. Although the report legitimately joins `OH_PATIENT` because it is a nominal document, the sex and age displayed for a specific examination should probably come from `PEX_SEX` and `PEX_AGE`, representing their values at examination time, rather than from the patient’s current `PAT_SEX` and `PAT_AGE`. Since the query currently uses SELECT *, this should be done carefully to avoid duplicate or ambiguous result-column names.

giuseppesorge:
@mwithi agreed on both reports. I'll alias the snapshot columns as `ADM_SEX AS PAT_SEX` / `ADM_AGE AS PAT_AGE` in `OH009_InPatientReport` so the existing Jasper fields keep working, and switch `rpt_base/patient_examination.jrxml` to `PEX_SEX` / `PEX_AGE` — being careful with its `SELECT *` so we don't introduce duplicate/ambiguous result columns.

One caveat, same as the other report edits on this PR: changing a `.jrxml` has no runtime effect until its committed `.jasper` is recompiled, and JR 7.0.6 can't recompile the legacy JR6-format files headlessly — they need the Jaspersoft Studio 7.0.2 migration (the same jasper track that OP-1403 / `AUTOCOMPILEJRXML` enables at startup). So I can prepare the query changes now, but the regenerated `.jasper` depend on that migration. How would you prefer to sequence it?

giuseppesorge:
Thanks @mwithi — you were right that editing the `.jrxml` alone wasn't enough, and it actually pointed to a deeper issue. OH loads the committed `.jasper` at runtime (`ReportLauncher` → `JRLoader.loadObject`), and JasperReports 7.0.6 can't recompile the shipped legacy JR6-format templates, so the query edits had **no runtime effect** until the `.jasper` was regenerated.

I regenerated them **headless**, without Jaspersoft Studio, by round-tripping each committed `.jasper` (which already holds a JR7 model) through `JRXmlWriter` to a JR7 `.jrxml` and recompiling with the pom's JR version. Text-render output is byte-identical to the previous `.jasper` on every report, so only the query changed.

- **The 4 `rpt_extra` HMIS reports** (`e44b2d1b`): migrated to JR7 and recompiled, so the snapshot query (`ADM_SEX`/`ADM_AGE`, no `OH_PATIENT` join) is now actually in the compiled report.
- **`OH009_InPatientReport`** (`512751c0`): as you suggested — `ADM_SEX AS PAT_SEX` / `ADM_AGE AS PAT_AGE`, keeping the `OH_PATIENT` join and the existing Jasper fields.

On **`patient_examination`**: I dug in and the premise doesn't quite hold — the report's own `SELECT *` isn't what displays sex/age. That comes from the shared subreport **`patient_clinical_sheet_patsubreport`**, which is used by **six** reports (admission/discharge/OPD charts, clinical sheet ver3, and patient_examination). It is keyed by `$P{patientID}` (not the examination id) and renders the **current patient header** from `OH_PATIENT`; its examination sub-select only pulls the latest `PEX_HEIGHT`/`PEX_WEIGHT`. So sourcing sex/age from `PEX_SEX`/`PEX_AGE` there would change the header of all six reports (a discharge chart has no "examination sex/age"), and it wouldn't even correspond to the specific examination being printed. Given `patient_examination` is a nominal clinical document that already shows the patient's identity, I'd propose leaving it as-is. Happy to add a dedicated examination-only element if you feel strongly, but it seems like low value for a nominal document — wdyt?

dbmalkovsky:
@giuseppesorge @mwithi I think this branch needs to be rebased on develop.  I'm getting a compile error in `org.isf.lab.gui.LabBrowser` line 202:
```
labs = labManager.getLaboratoryForPrint(typeSelected, dateFrom.getDateStartOfDay(), dateTo.getDateEndOfDay());
```
It indicates there should be 4 parameters but there are only 3.

This PR is dated July 17 but on July 27  @giuseppesorge updated the code to be (using 4 parameters):
```
labs = labManager.getLaboratoryForPrint(typeSelected, dateFrom.getDateStartOfDay(), dateTo.getDateEndOfDay(), null);
```
So, `develop` has changes not reflect in this PR.

dbmalkovsky:
FWIW, Once I checked this code out and `rebased` it onto `develop` things worked successfully.

giuseppesorge:
Thanks @dbmalkovsky — your diagnosis was exactly right, and thanks for taking the time to check the branch out and confirm the rebase fixed it.

The branch was based on `develop` from before #2216 (OP-1065), which changed the `getLaboratoryForPrint` call site in `LabBrowser` to the four-argument form. Nothing in this PR touches `LabBrowser`, so it was purely a stale base — sorry for the wasted round.

Rebased onto `develop` (`512751c0` → `a780e3d7`), no conflicts; the three commits are unchanged in content and the PR still touches only the five report files.

Verified after the rebase, against CORE #1602 on a demo database built from its `create_all_demo.sql` (MariaDB 10.6):

- `mvn compile` clean.
- Ran **HMIS 108 – In Patient (diagnosis In)** for 10/2022 in the GUI: the recompiled JR7 `.jasper` renders, and the cells match the database exactly — Malaria 5/8/1, Tuberculosis 3/1, Typhoid Fever 1/1, UTI 6/1.
- Compared every report's query, old vs new, on the same data:

| report | rows | old total | new total |
|---|---|---|---|
| hmis033 under 5 | 48 | 161 | **152** |
| hmis033 over 5 | 48 | 389 | **398** |
| hmis108 diagnosisIn | 110 | 117 | 117 (identical row by row) |
| hmis108 diagnosisOut | 110 | 117 | 117 (identical row by row) |

The 9 admissions that move between the hmis033 bands are the point of the change, not a regression: the old queries read `OH_PATIENT.PAT_AGE`, a stored column, while the new ones read the age at the admission date. In the demo dataset `PAT_AGE` is out of sync with `PAT_BDATE` by two years on 427 of 460 admissions — e.g. admission 183 is a patient born 2017-09-04 admitted 2022-10-25, so 5 at admission, while `PAT_AGE` says 3. `ADM_AGE` matches the value computed from `PAT_BDATE` on **460 of 460** admissions.

Same effect on OH009: sex is identical on all 147 rows, age changes on all of them, always to the value computed from the birth date at the admission date.

giuseppesorge:
@mwithi — coming back to your review. The first half is done (`512751c0`): `OH009_InPatientReport` now selects `ADM_SEX AS PAT_SEX` / `ADM_AGE AS PAT_AGE`, keeping the existing Jasper fields. On demo data the sex comes out identical on all 147 rows and the age moves to the snapshot value, as you expected.

On the examination side you were right that it shows the patient's current values — I just got lost on where the change belongs, so I would rather ask than pick for you.

In `patient_examination.jrxml` itself the `PAT_SEX` and `PAT_AGE` fields are declared but never rendered; there is no `$F{...}` expression using them, they look like leftovers of the `SELECT *`. The values that actually appear on the printout come from the `patient_clinical_sheet_patsubreport` subreport, which is also used by `patient_clinical_sheet_ver3`, `patient_adm_chart`, `patient_opd_chart` and `patient_dis_chart`. It receives only `patientID` and queries `OH_PATIENT` by `PAT_ID`, so as it stands it has no way of knowing which examination it is being rendered for.

That leaves a few directions, and I do not have a strong view on which fits the project best:

- pass the examination id down as a new parameter, falling back to the patient's current values when it is absent, since the other four parents have none to pass;
- add an examination-specific variant of the header subreport and leave the shared one untouched;
- or keep this PR to the admission reports and handle the examination printout separately.

One thing that might help weigh the third: as far as I can tell no report in the repository reads `PEX_SEX`/`PEX_AGE` today, so nothing is currently displaying a wrong value on that side — while the admission half is what the five reports here use for the aggregate statistics. That said, you know the reporting side far better than I do, so if you would rather see it closed in one go I am happy to implement either of the first two here.

Whichever way, the file is still in the JR6 format, so it needs the same `.jasper` round-trip as the other five.

mwithi (approved):
LGTM. Thanks a lot!

PR #2239, opened 2026-07-17
OP-1426 Keep the inventory window open when a save-on-close fails

giuseppesorge:
## OP-1426

Closing an inventory edit window with unsaved changes offers to save them, but if the save failed validation the window closed anyway and the entered data was lost.

### Root cause
In both `InventoryEdit.getCloseButton()` and `InventoryWardEdit.getCloseButton()` the *Yes* branch ran `saveButton.doClick()` and then called `dispose()` **unconditionally** — the save outcome was never checked, so any validation failure (e.g. a missing inventory reference) still ended in the window being disposed.

### Fix
Track whether the save actually succeeded with a `saveSucceeded` flag: reset to `false` at the start of the save action and set to `true` only at the successful end of the new/update paths (a validation failure throws an `OHServiceException` caught before that point). The close button now disposes the window only when `saveSucceeded` is true. Applied to both InventoryEdit and InventoryWardEdit.

### Steps to reproduce (now fixed)
Pharmacy → Inventory → New → *Add medicals by* → Select → All → OK (populates the table) → leave *Inventory's reference* empty → Close → Yes → the save fails with "You must enter a reference." After dismissing the error the window now stays open with the rows intact, so the reference can be entered and the inventory saved.

Compiles; copyright year bumped to 2026.

### DIFF
--- a/src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
+++ b/src/main/java/org/isf/medicalinventory/gui/InventoryEdit.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2025 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
+ * Copyright © 2006-2026 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
  *
  * Open Hospital is a free and open source software for healthcare data management.
  *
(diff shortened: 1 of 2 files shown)


mwithi:
Fix applied directly on the release branch https://github.com/informatici/openhospital-gui/commit/8185641223c61182054104f27c54f74b403988e1
Thanks a lot!

PR #2244, opened 2026-07-28
OP-1440 Make the macOS launcher start Open Hospital on a first run

giuseppesorge:
Five defects in `ohmac.sh` that form a single chain: the macOS package does not start on a clean installation, and each one is only visible once the previous is out of the way.

Reproduced and verified on a freshly extracted `OpenHospital-v1.15.1-multiarch-client.zip`, letting the launcher download the JRE itself, with nothing changed but this script.

| | what happens today |
|---|---|
| [OP-1448](https://openhospital.atlassian.net/browse/OP-1448) | `JAVA_BIN` is set only when the JRE is already on disk, so on a first run the launch command begins with `-client` and the shell tries to run that as a program. Every first installation fails; a second run works. |
| [OP-1447](https://openhospital.atlassian.net/browse/OP-1447) | Four of the eleven configuration files are generated. `sms.properties` and `telemetry.properties` are `@PropertySource` in the core, so the application context fails to build and the GUI dies during startup. |
| [OP-1440](https://openhospital.atlassian.net/browse/OP-1440) | The connection test runs `mysql`, which the package does not ship and macOS does not provide, authenticates with the operating system user and passes the root password, then `exit 2`. It can never pass. |
| [OP-1446](https://openhospital.atlassian.net/browse/OP-1446) | The output is redirected into a directory that is never created. A redirection to a missing directory makes the shell skip the command, so the JVM is never launched and the only message is "An error occurred while starting Open Hospital". Three calls to `shutdown_database`, which exists only in `oh.sh`, are corrected to `stop_db`. |
| [OP-1441](https://openhospital.atlassian.net/browse/OP-1441) | `rsc/images` is missing from the classpath, so `MainMenu.getLogoPanel` gets a null resource for `logo_menu_vert.png` and the main menu dies with an NPE right after a successful login. `oh.sh`, `oh.bat` and `oh.ps1` all carry that entry. |

One commit per defect, in the order a first run meets them, so each can be read on its own.

## Verification

With all five applied, from a freshly extracted package: the launcher downloads the JRE, starts, login works, the main menu opens, and the application log has no exceptions.

The GUI regression suite runs green on this branch, including the case that covers the classpath entry — it drives the real application and asserts the main menu opens with the classpath read from `ohmac.sh` itself, so it will catch a future removal.

## Notes for review

`copy_config_file` deliberately writes only when a file is missing, unlike its counterpart in `oh.sh`. `WRITE_CONFIG_FILES` is fixed to `"on"` in this script with no option to change it, so copying on every run would replace what a site configured — vital sign ranges, SMS gateway credentials — and a second run would overwrite the backup the first had just made.

Three further defects found in the same file while working on these are filed separately and are **not** touched here: [OP-1449](https://openhospital.atlassian.net/browse/OP-1449) (the API configuration template is destroyed by a `sed` on the `.dist` file), [OP-1450](https://openhospital.atlassian.net/browse/OP-1450) (`database.properties` is ignored and overwritten, so a remote database cannot be configured on macOS) and [OP-1451](https://openhospital.atlassian.net/browse/OP-1451) (four functions are called that only exist in `oh.sh`).

[OP-1448]: https://openhospital.atlassian.net/browse/OP-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1447]: https://openhospital.atlassian.net/browse/OP-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1440]: https://openhospital.atlassian.net/browse/OP-1440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1446]: https://openhospital.atlassian.net/browse/OP-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ohmac.sh
+++ b/ohmac.sh
@@ -45,9 +45,16 @@ OH_API_JAR="openhospital-api-0.1.0.jar"
 # OH configuration files
 OH_SETTINGS="settings.properties"
 DATABASE_SETTINGS="database.properties"
+EXAMINATION_SETTINGS="examination.properties"
 IMAGING_SETTINGS="dicom.properties"
 LOG4J_SETTINGS="log4j2-spring.properties"
+PRINTER_SETTINGS="txtPrinter.properties"
+SMS_SETTINGS="sms.properties"
+TELEMETRY_SETTINGS="telemetry.properties"
+XMPP_SETTINGS="xmpp.properties"
 API_SETTINGS="application.properties"
+CRED_SETTINGS="default_credentials.properties"
+DEMO_CRED_SETTINGS="default_demo_credentials.properties"
 HELP_FILE="OH-readme.txt"
 
 # logging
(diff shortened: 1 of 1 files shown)


mwithi (approved):
LGTM. Thanks a lot!

PR #2245, opened 2026-07-28
OP-1443 Make the portable launcher survive an imperfect first run

giuseppesorge:
Two defects in the portable path of `oh.sh`, both about a first run that does not go perfectly. Reproduced and verified in a Linux x86_64 container running the portable package, with a before and after for each.

## [OP-1443](https://openhospital.atlassian.net/browse/OP-1443) — the launcher hangs waiting for the database

The wait for the database port used `nc`, which the package does not ship and many minimal installations do not have. There the probe fails every time, so the loop never ends:

```
./oh.sh: line 804: nc: command not found
./oh.sh: line 804: nc: command not found
... once a second, forever
```

The database is up and serving throughout. Nothing on screen suggests what is wrong.

The port is now probed with bash's own `/dev/tcp` redirection, so nothing outside the package is needed, and both waits give up after `DATABASE_WAIT_TIMEOUT` seconds instead of spinning — the start-up one with an error, the shutdown one with a warning rather than the "stopped!" that would not be true.

## [OP-1444](https://openhospital.atlassian.net/browse/OP-1444) — an interrupted installation is never recognised

Whether OH was installed was decided by the existence of the data directory, which `initialize_database` creates as its very first step. An installation interrupted at any point afterwards therefore looks complete on the next run:

```
OH database found!
Starting MariaDB server...
Error: can't connect to database! Exiting.
```

and it does that on every subsequent run. Nothing says the first attempt never finished, so the only way out is knowing to delete the directory by hand.

The check now looks for the directory the database engine creates for the schema itself, which exists only once the database has actually been created.

## Verification

| | before | after |
|---|---|---|
| no `nc` available | loop with no end, database already up | `MariaDB server started!`, server ready |
| interrupted installation | `OH database found!` → `can't connect to database` | the unfinished installation is named, with how to recover |
| complete installation | starts | starts |

The last row is the one that matters most for OP-1444, and it holds for installations made by earlier versions too: the criterion is a directory the engine has always created.

## Note for review

The schema directory is matched without regard to case. The shipped `my.cnf` sets `lower_case_table_names`, so a database named `MyHospital` is stored in a directory called `myhospital` while `DATA_DIR` keeps the name as the user typed it — comparing the two literally would have declared a perfectly healthy installation unfinished, and the message would have invited the user to delete it.

[OP-1443]: https://openhospital.atlassian.net/browse/OP-1443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1444]: https://openhospital.atlassian.net/browse/OP-1444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oh.ps1
+++ b/oh.ps1
@@ -137,6 +137,9 @@ $script:OH_DIR="."
 $script:OH_DOC_DIR="doc"
 $script:CONF_DIR="data/conf"
 $script:DATA_DIR="data/db"
+
+# seconds to wait for the database to start listening, or to release its port on shutdown
+$script:DATABASE_WAIT_TIMEOUT=90
 $script:PHOTO_DIR="data/photo"
 $script:BACKUP_DIR="data/dump"
 $script:LOG_DIR="data/log"
@@ -893,18 +896,38 @@ function start_database {
 
 	Write-Host "Starting $MYSQL_NAME server... "
 	try {
-		Start-Process -FilePath "$OH_PATH\$MYSQL_DIR\bin\mysqld.exe" -ArgumentList ("--defaults-file=`"$OH_PATH\$CONF_DIR\$MYSQL_CONF_FILE`" --tmpdir=`"$OH_PATH\$TMP_DIR`" --standalone") -NoNewWindow -RedirectStandardOutput "$LOG_DIR/$LOG_FILE" -RedirectStandardError "$LOG_DIR/$LOG_FILE_ERR"
-		Start-Sleep -Seconds 2
+		$script:DATABASE_PROCESS = Start-Process -PassThru -FilePath "$OH_PATH\$MYSQL_DIR\bin\mysqld.exe" -ArgumentList ("--defaults-file=`"$OH_PATH\$CONF_DIR\$MYSQL_CONF_FILE`" --tmpdir=`"$OH_PATH\$TMP_DIR`" --standalone") -NoNewWindow -RedirectStandardOutput "$LOG_DIR/$LOG_FILE" -RedirectStandardError "$LOG_DIR/$LOG_FILE_ERR"
 	}
 	catch {
 		Write-Host "Error: $MYSQL_NAME server not started! Exiting." -ForegroundColor Red
 		Read-Host; exit 2
 	}
 
-	# wait till the MariaDB/MySQL socket file is created -> TO BE IMPLEMENTED
-	# while ( -e $OH_PATH/$MYSQL_SOCKET ); do sleep 1; done
-	# # Wait till the MariaDB/MySQL tcp port is open
-	# until nc -z $DATABASE_SERVER $DATABASE_PORT; do sleep 1; done
+	# Wait till the MariaDB/MySQL tcp port is open, instead of assuming two seconds are enough: the
+	# next thing this script does is run a mysql client, and on a slow machine, or on a start that
+	# has to build the system tables or recover after an unclean shutdown, two seconds are not.
+	#
+	# A start that has already failed is not waited out either - once mysqld has exited, the port
+	# will never open and there is nothing left to wait for. That is the common failure and it is
+	# reported at once; the timeout covers the rarer server that runs but does not get to listening.
+	$WAITED = 0
+	while ( !(database_port_open) ) {
+		if ( $script:DATABASE_PROCESS -And $script:DATABASE_PROCESS.HasExited ) {
+			Write-Host "Error: $MYSQL_NAME server exited before it started listening on $DATABASE_SERVER port $DATABASE_PORT." -ForegroundColor Red
+			Write-Host "See $LOG_DIR/$LOG_FILE_ERR for the reason. Exiting." -ForegroundColor Red
+			Read-Host; exit 2
+		}
+		if ( $WAITED -ge $DATABASE_WAIT_TIMEOUT ) {
+			Write-Host "Error: $MYSQL_NAME server is not listening on $DATABASE_SERVER port $DATABASE_PORT after $DATABASE_WAIT_TIMEOUT seconds." -ForegroundColor Red
+			Write-Host "See $LOG_DIR/$LOG_FILE_ERR for the reason. Exiting." -ForegroundColor Red
+			Read-Host; exit 2
+		}
+		if ( ($WAITED -gt 0) -And ($WAITED % 10 -eq 0) ) {
+			Write-Host "Still waiting for $MYSQL_NAME to listen on $DATABASE_SERVER port $DATABASE_PORT ($WAITED s)..."
+		}
+		Start-Sleep -Seconds 1
+		$WAITED++
+	}
 
 	Write-Host "$MYSQL_NAME server started!"
 }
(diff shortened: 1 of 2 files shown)


mwithi:
Could these fixes also be applied to the equivalent portable database paths in `oh.ps1`? It currently relies on fixed two-second sleeps without checking whether the database port opened/closed, and considers the installation complete as soon as DATA_DIR exists.

`ohmac.sh` would also benefit from the bounded start/stop readiness check after brew services, although the incomplete-datadir fix does not apply there because it uses Homebrew’s database storage and already checks the OH_VERSION table.

Also, could DATABASE_WAIT_TIMEOUT be reduced from 90 to 5 seconds? Since the port is checked once per second, five seconds should still accommodate a normal local startup while avoiding an unnecessarily long wait when startup or shutdown has failed.

giuseppesorge:
@mwithi done in `764dd233`, plus one part that landed on the other PR.

**`oh.ps1`** had both defects, and neither was covered before: the two fixed `Start-Sleep -Seconds 2` (with the readiness check sitting there as a `TO BE IMPLEMENTED` comment) are now bounded waits on the TCP port, and the installed/not-installed test now looks for the schema directory the engine creates inside the data directory, exactly as `oh.sh` does. The case handling `oh.sh` needs has no equivalent here — Windows filesystems are case-insensitive, so a plain `Test-Path` is enough.

**`ohmac.sh`** has the bounded start/stop check after `brew services`, but it went on **#2247** rather than here: this branch does not touch `ohmac.sh`, and two open PRs editing the same file would leave a conflict for whichever is merged second. Commit `ebfa014e` there.

---

**On `DATABASE_WAIT_TIMEOUT` 90 → 5, I would rather not, and I think there is a better answer to what you are after.**

What the long timeout actually costs is a wait when *startup has failed* — and in that case there is no need to wait at all. `mysqld_safe` stays alive for as long as the server does, so once it has exited the port will never open and there is nothing left to wait for. That is the common failure — port already taken, a data directory the server will not accept, a bad config — and it now reports in about a second, whatever the timeout is. `oh.ps1` does the same through the process handle from `Start-Process -PassThru`. Both also print a line every ten seconds, so a slow start no longer looks like the hang these commits replaced.

That leaves the timeout covering only the rarer case of a server that is running but has not got to listening yet — where waiting is the correct behaviour, and where 5 seconds would turn a start that would have succeeded into a hard `exit 2`. I measured it rather than guessed: MariaDB 10.6 on a fresh data directory, probed from inside the container so no port-forwarder answers early, on an Apple Silicon machine with the image already local:

```
run 1: port open at 2706 ms | query accepted at 2742 ms
run 2: port open at 2689 ms | query accepted at 2716 ms
run 3: port open at 2678 ms | query accepted at 2707 ms
```

~2.7 s is the *best* case: fast machine, SSD, no data to recover, no OH schema to import. A 5-second budget is already 54% spent there, and an older machine, a spinning disk, or a start recovering after an unclean shutdown goes past it easily — and the user gets "not listening, exiting" on a database that was about to come up.

So: the fast path you wanted is in, without the false-failure risk. If you still prefer 5 after this, say so and I will change the one line — it is your call, I just did not want to make it silently.

---

**What I checked.** `oh.sh`: `bash -n`, and the wait helpers exercised against a real listener — port closed → exits on timeout with code 2; listener up → returns immediately; stop-wait with the listener still up → warns and returns; listener closed → returns at once.

`oh.ps1`: parsed clean with the PowerShell parser (7.4.2, in a container — `pwsh` is not available on this machine), and `database_port_open` exercised there against a real socket: closed port → `False` in 167 ms with no exception escaping, open port → `True` in 4 ms, after the listener stops → `False`, and the wait loop leaves on the timeout. The guard matters: a refused connection makes `Task.Wait` throw rather than return false, which the existing port-scan loop above it gets away with only because PowerShell swallows the method exception and leaves the variable unset.

**What I could not check: `oh.ps1` on Windows.** The logic is exercised on PowerShell 7.4 on Linux, but Windows PowerShell 5.1, the real `mysqld.exe` and the `HasExited` path are not. Worth a run on a Windows box before this merges.

giuseppesorge:
Closing the gap I left open above: **`oh.ps1` is now checked on real Windows**, not only parsed on PowerShell 7 on Linux.

I ran it on a Windows runner (image `windows-2025-vs2026`), under **Windows PowerShell 5.1** — the one the launcher actually ships for — and again under PowerShell 7, against the **`mysqld.exe` from the released `OpenHospital-v1.15.0-windows_x86_64-portable` package** rather than a stand-in. The checks execute the real lines taken out of `oh.ps1`, not a copy of them.

```
=== PowerShell 5.1.26100.32995 (Desktop) ===

--- 1. oh.ps1 parses ---
  OK - no syntax errors

--- 3. database_port_open, taken from oh.ps1 ---
  OK   closed port -> False, no exception escaped
  OK   listening port -> True
  OK   listener stopped -> False

--- 4. the real mysqld.exe: a start that fails is not waited out ---
  using ...\pkg\OpenHospital-v1.15.0-windows_x86_64-portable\mariadb-10.6.23-winx64\bin\mysqld.exe
  OK   HasExited caught the dead server after 1.0s, not at the 90s timeout
  mysqld said: Could not open required defaults file: ...\does-not-exist.cnf /
               Fatal error in defaults handling. Program aborted

=== failures: 0 ===
```

Identical under PowerShell 7.6.3.

The fourth one is the point of the change: `mysqld.exe` was started with a configuration it refuses, and the wait loop returned in **1 second** through `Start-Process -PassThru` / `.HasExited`, instead of sitting out the 90-second timeout. That is the answer to the question about shortening `DATABASE_WAIT_TIMEOUT` — the case that made the wait feel long now ends at once, whatever the timeout is set to, and a slow-but-succeeding start is still given room.

The `database_port_open` guard matters on 5.1 too: a refused connection makes `Task.Wait` throw, and the closed-port case returns `False` with nothing escaping.

Run on a branch of my fork, not part of this PR. Two things stay unverified and I would rather say so: a full portable start, which needs an initialised data directory, and anything to do with the Swing GUI on Windows.

giuseppesorge:
Following up on the Windows results above: the two defects this PR fixes are Linux ones, so I have now checked them **on Linux**, against the released package rather than by reading the script.

Ubuntu 24.04, bash 5.2.21, `OpenHospital-v1.15.0-linux_x86_64-portable`, running the lines taken out of `oh.sh`:

```
=== Linux x86_64 | bash 5.2.21(1)-release ===

--- 2. database_port_open, taken from oh.sh, with no netcat on PATH ---
  (netcat is present on this runner: hiding it)
  OK   closed port    -> false, with no netcat reachable
  OK   listening port -> true,  with no netcat reachable
  OK   listener stopped -> false

--- 3. an unfinished installation is told apart from a finished one ---
  expression: find ./"$DATA_DIR" -mindepth 1 -maxdepth 1 -type d -iname "$DATABASE_NAME" -print -quit
  OK   datadir with no schema          -> unfinished
  OK   datadir with the schema         -> finished
  OK   schema dir in a different case  -> finished
  OK   a datadir whose own name matches -> still unfinished

--- 4. the real mysqld from the released package ---
  using ./pkg/.../mariadb-10.6.23-linux-systemd-x86_64/bin/mariadbd
  OK   the dead server is seen after 1s, not at the 90s timeout
```

Three things worth calling out.

**The netcat check is now real.** The probe runs with an emptied `PATH`, and it also asserts that `nc` is genuinely unreachable before probing — so the check cannot quietly turn into a no-op. That is OP-1443 reproduced on the platform it happens on, rather than argued from the code.

**The case row matters in practice.** OH ships `lower_case_table_names=1`, so a database named `MyHospital` is stored in a directory called `myhospital` while `DATA_DIR` keeps the name the user typed. Without `-iname` a perfectly good installation would be declared unfinished, and the message tells the user to delete their data.

**The last row guards a regression I nearly shipped.** Without `-mindepth 1`, `find` matches the starting directory itself, so every fresh installation would look finished. The check pins that.

Also verified on **macOS** (the `ohmac.sh` side, on `macos-latest`) and on **Windows**, so all three platforms this launcher family ships on are now covered by a run rather than by inference. The checks live on a branch of my fork, not in this PR.

Still not covered, and I would rather name it: a full portable start, which needs an initialised data directory, and anything involving the Swing GUI.

mwithi (approved):
LGTM. Thanks a lot!

PR #2246, opened 2026-07-28
OP-1442 Open the patient form where the webcam native is missing

giuseppesorge:
Fixes [OP-1442](https://openhospital.atlassian.net/browse/OP-1442).

Enabling the video module states that the user wants a webcam, not that the platform can provide one. The capture driver loads a native library through BridJ, which ships for a limited set of architectures; where it is absent `Webcam.getDefault()` throws, the failure escapes the `PatientPhotoPanel` constructor and **no patient can be opened at all**:

```
java.lang.RuntimeException: Failed to initialize BridJ
Caused by: java.lang.UnsatisfiedLinkError: ... libbridj.dylib:
    (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
```

`VIDEOMODULEENABLED=yes` is what the packages ship, so this is the configuration users actually get. OP-1417 had already made the *disabled* case harmless; this is the enabled one.

The lookup now falls back to no webcam, which is the path already taken when the video module is off: the photo is attached from a file instead. The reason is logged once, so an installation that expected a working webcam can still be diagnosed.

## Not only macOS

The bundled `bridj-0.7.0.jar` carries natives for `darwin_universal`, `linux_x64`, `linux_x86`, `linux_armhf`, `win32` and `win64`. There is **no `linux_arm64`**, so a 64-bit ARM machine — a Raspberry Pi, which is not an exotic deployment for Open Hospital — fails in exactly the same way. It is visible today on Apple Silicon because that is where the mismatch was hit.

## Verification

Driven against the running application, before and after, with the video module enabled as shipped:

* before — the patient form does not open, six native errors in the log;
* after — the form opens, and the only entry left is the single warning explaining why the photo booth is not offered.

The case is part of the GUI regression suite, so a future change that lets the exception escape again will be caught.

[OP-1442]: https://openhospital.atlassian.net/browse/OP-1442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/isf/video/gui/PatientPhotoPanel.java
+++ b/src/main/java/org/isf/video/gui/PatientPhotoPanel.java
@@ -153,10 +153,10 @@ public class PatientPhotoPanel extends JPanel {
 				}
 			});
 
-			// only touch the webcam stack when the video module is enabled; Webcam.getDefault() loads a native library
+			// only touch the webcam stack when the video module is enabled; the lookup loads a native library
 			// (BridJ) that is not available on every platform (e.g. Apple Silicon), and calling it unconditionally
 			// would throw an UnsatisfiedLinkError and break the whole patient form even with the video module off
-			final Webcam webcam = GeneralData.VIDEOMODULEENABLED ? defaultWebcam() : null;
+			final Webcam webcam = GeneralData.VIDEOMODULEENABLED ? WebcamProvider.getDefault() : null;
 
 			if (webcam != null) {
 				JButton jGetPhotoButton = new JButton(MessageBundle.getMessage("angal.patientphoto.newphoto.btn"));
@@ -190,25 +190,6 @@ public class PatientPhotoPanel extends JPanel {
 		}
 
 		add(jPhotoPanel);
-	}
-
-	
-	/**
-	 * The default webcam, or {@code null} when none can be used.
-	 * <p>
-	 * Enabling the video module only states that the user wants a webcam, not that the platform can provide one: the
-	 * capture driver loads a native library (BridJ) that ships for a limited set of architectures, so on the others the
-	 * lookup fails at run time. Without this the failure escapes the constructor and no patient can be opened at all,
-	 * which is a far worse outcome than losing the photo booth. Falling back to {@code null} reuses the path already
-	 * taken when the video module is off, where the photo is attached from a file instead.
-	 */
-	private static Webcam defaultWebcam() {
-		try {
-			return Webcam.getDefault();
-		} catch (RuntimeException | LinkageError e) {
-			LOGGER.warn("No usable webcam on this platform: taking the photo from a file instead.", e);
-			return null;
-		}
 	}
 
 	private JPanel setMyBorder(JPanel c, String title) {
--- a/src/main/java/org/isf/video/gui/PatientPhotoPanel.java
+++ b/src/main/java/org/isf/video/gui/PatientPhotoPanel.java
@@ -1,6 +1,6 @@
 /*
  * Open Hospital (www.open-hospital.org)
- * Copyright © 2006-2023 Informatici Senza Frontiere (info@informaticisenzafrontiere.org)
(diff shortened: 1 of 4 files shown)


dbmalkovsky (commented):
While this seems like it will work, the downside is that every time the Patient is opened the check is made again.  Would it be better to do the check once in GenerakData.java and if the platform can't provide a webcam,  set GeneralData.VIDEOMODULEENABLED to false. If there is a webcam, use a new GeneralData variable that was the value of "Webcam.getDefault()'.  wdyt?

mwithi:
> While this seems like it will work, the downside is that every time the Patient is opened the check is made again. Would it be better to do the check once in GenerakData.java and if the platform can't provide a webcam, set GeneralData.VIDEOMODULEENABLED to false. If there is a webcam, use a new GeneralData variable that was the value of "Webcam.getDefault()'. wdyt?

I agree that repeatedly calling `Webcam.getDefault()` for every opened patient is unnecessary, especially when native initialization has already failed permanently.
However, I would avoid storing a Webcam instance in GeneralData or changing `VIDEOMODULEENABLED` to false. GeneralData belongs to the core module, while Webcam is a GUI dependency; also, the flag represents the user’s configuration, not the runtime availability of the device on the client.
Could the detection result instead be cached in a GUI-level (e.g.) `WebcamProvider.java` (or statically in PatientPhotoPanel) and reused by `PhotoboothComponentImpl` as well? Ideally, a permanent native/BridJ failure should be cached and logged once, while a simple `null` result could remain retryable so a webcam connected after startup can still be detected.

giuseppesorge:
@dbmalkovsky @mwithi done in `e6ab6d8e`, taking @mwithi's shape.

The detection now lives in a GUI-level `WebcamProvider` rather than in `GeneralData`, for the two reasons given: `Webcam` is a GUI dependency and `GeneralData` is core, and `VIDEOMODULEENABLED` records what the user asked for, not what the machine in front of them can deliver. Writing the runtime answer back into the setting would also be destructive in a client installation — one machine without a camera would turn the photo booth off for everyone sharing that configuration file.

**Only a driver failure is remembered.** That one is permanent for the life of the process: the native library will not appear later, and once its class has failed to initialize, every further call fails again. An ordinary `null` — no device connected — is not remembered, so a webcam plugged in after startup is still picked up, as you asked.

`PhotoboothComponentImpl` goes through the provider as well, for both `getDefault()` and `getWebcams()`.

**Measured on the platform that actually fails** (Apple Silicon, where BridJ ships no arm64 native — `libbridj.dylib` is x86_64 only), reading the cache flag by reflection:

```
prima:            driverUnavailable=false
getDefault() #1 -> null | 23 ms | driverUnavailable=true
getDefault() #2 -> null |  0 ms | driverUnavailable=true
getDefault() #3 -> null |  0 ms | driverUnavailable=true
getWebcams()  -> []     |  0 ms
```

So the first patient form pays the failure, and no one after it does.

Worth recording, since it also confirms the premise: `Webcam.getDefault()` here throws `WebcamException` — a `RuntimeException`, wrapping the BridJ `UnsatisfiedLinkError` — and it throws on **every** call, not only the first. It is not a case of the library degrading to `null` by itself.

`WebcamProviderTest` pins the part that matters and is portable: the lookup never escapes, and repeating it keeps answering. It exercises the failure path here and the ordinary path where the driver loads.

One thing I left alone, so it is a conscious omission rather than an oversight: `PhotoboothComponentImpl` dereferences `webcam.getDevice()` straight after the lookup, so a `null` there would still NPE. It is unreachable today — the dialog only opens from the button, which only exists when a webcam was found — and guarding it properly means deciding what the dialog should do when the device disappears between the click and the constructor. Happy to take it here if you would rather not leave it, otherwise it is a separate issue.

mwithi (approved):
LGTM. Thanks a lot!

PR #2247, opened 2026-07-28
OP-1449 Fix the remaining macOS launcher defects: API configuration, database configuration, missing functions

giuseppesorge:
The three remaining defects of `ohmac.sh`, found while working on the ones in #2244 and deliberately kept out of it. This branch stands on its own; the eight commits across both were verified together on a freshly extracted package.

## [OP-1449](https://openhospital.atlassian.net/browse/OP-1449) — the API configuration is never written

`write_api_config_file` builds its paths from `$OHDIR`, which this script does not define — the name is `OH_DIR` — so every path points at `/rsc`, outside the installation:

```
Writing OH API configuration file -> application.properties...
sed: /rsc/application.properties.dist: No such file or directory
cp: /rsc/application.properties.dist: No such file or directory
```

`application.properties` is never created and the API server starts with no configuration of its own.

Two more defects sit in the same seven lines: the guard reads `[ ! -f SET_FILE ]` without the `$`, so it is always true, and the substitutions are applied **in place to the template** with `sed -i` before copying it over the target. That last one is harmless today only because the first sends it nowhere — correcting the path alone would consume the placeholders in the shipped `application.properties.dist`, of which the package holds no second copy. The template is now read and the copy written, as `oh.sh` does.

## [OP-1450](https://openhospital.atlassian.net/browse/OP-1450) — a macOS installation cannot keep its own database

`read_settings` reads only `version.properties`, so the connection details stay at the values hardcoded in the script whatever `database.properties` says — and `write_config_files` then writes those values back over the file on every run, because `WRITE_CONFIG_FILES` is fixed to `"on"` here with no way to change it. A macOS installation therefore always talks to `isf@localhost:3306/oh`, and CLIENT mode against a hospital server — the reason that mode exists — cannot be configured at all.

The file is now read as `oh.sh` reads it, and it is written only when it is not there. Just that one: it holds a decision of whoever installed OH, unlike the files around it, which the launcher derives from the mode, language and paths it was given and keeps regenerating exactly as before — so nothing else about a run changes. The packages ship only `.dist` templates, so a first run still creates the whole set.

## [OP-1451](https://openhospital.atlassian.net/browse/OP-1451) — four functions that do not exist

Turning the API server on reaches one of them: `start_api_server: command not found`, the launcher says the API server is enabled and starts nothing. It is now written for what the macOS packages actually contain — a self-contained Spring Boot artifact, not the Tomcat layout `oh.sh` starts — and says so plainly when the package carries no API at all.

The other three calls are unreachable today because the options that would enable them are commented out of `parse_user_input` and their variables never assigned. That is a defect in itself: the menu offers `-E    toggle EXPERT MODE - show advanced options` and pressing it does nothing, while the advanced options themselves are implemented. `script_menu_advanced` now lists the four this script really has, so enabling `-E` becomes a decision rather than a promise, and `start_ui` opens a browser the way macOS does. The demo branch used `DEMO_DATABASE` and `DB_DEMO` without defining either; both are now set, and the `set_db_name` call is dropped rather than ported, since it computes a portable data directory this launcher does not use.

## Verification

On a freshly extracted package, with these three and the five in #2244 applied together:

* first run — the JRE is downloaded and used, the whole set of configuration files is generated, the GUI starts and the login window opens;
* second run after editing `database.properties` — `Reading database settings file...` and the file is left byte-identical;
* `start_api_server` with no artifact, with an artifact but no configuration, and with both — a clear message, a clear message, and the launcher command that runs the API server;
* no function is called that the script does not define;
* an installation that works today keeps working: the only file whose handling changes is `database.properties`, and the API server path can only add a warning where there used to be a `command not found`.

## Note on the two descriptions that changed

OP-1449 and OP-1451 were first filed with claims that did not survive being reproduced properly — a destroyed template that is in fact untouched, and four failing calls of which only one can fire. Both issues carry the correction and the evidence; the code here matches what actually happens.

[OP-1449]: https://openhospital.atlassian.net/browse/OP-1449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OP-1450]: https://openhospital.atlassian.net/browse/OP-1450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oh.sh
+++ b/oh.sh
@@ -1013,13 +1013,7 @@ function write_config_files {
 	if [ "$WRITE_CONFIG_FILES" = "on" ] || [ ! -f ./$OH_DIR/rsc/$OH_SETTINGS ]; then
 		[ -f ./$OH_DIR/rsc/$OH_SETTINGS ] && mv -f ./$OH_DIR/rsc/$OH_SETTINGS ./$OH_DIR/rsc/$OH_SETTINGS.old
 		echo "Writing OH configuration file -> $OH_SETTINGS..."
-		# The three interface expressions are anchored at the start of the line, and name the key on
-		# both sides of the substitution:
-		#  - the GUI_INTERFACE replacement used to expand the variable on the left of the '=' as well,
-		#    so it wrote "on=on" (or "off=off") and the key itself was lost from the settings file;
-		#  - UI_INTERFACE is a substring of GUI_INTERFACE, so an unanchored UI_INTERFACE expression
-		#    matches inside the GUI_INTERFACE line the previous expression has just produced and
-		#    rewrites it. Anchoring is what keeps the two apart.
+		# anchored: UI_INTERFACE is a substring of GUI_INTERFACE
 		sed -e "s/OH_MODE/$OH_MODE/g" -e "s/OH_LANGUAGE/$OH_LANGUAGE/g" -e "s&OH_DOC_DIR&../$OH_DOC_DIR&g" \
 		-e "s/DEMODATA=off/"DEMODATA=$DEMO_DATA"/g" -e "s/YES_OR_NO/$OH_SINGLE_USER/g" \
 		-e "s/PHOTO_DIR/$PHOTO_DIR_ESCAPED/g" \
--- a/oh.sh
+++ b/oh.sh
@@ -1013,10 +1013,19 @@ function write_config_files {
 	if [ "$WRITE_CONFIG_FILES" = "on" ] || [ ! -f ./$OH_DIR/rsc/$OH_SETTINGS ]; then
 		[ -f ./$OH_DIR/rsc/$OH_SETTINGS ] && mv -f ./$OH_DIR/rsc/$OH_SETTINGS ./$OH_DIR/rsc/$OH_SETTINGS.old
 		echo "Writing OH configuration file -> $OH_SETTINGS..."
+		# The three interface expressions are anchored at the start of the line, and name the key on
+		# both sides of the substitution:
+		#  - the GUI_INTERFACE replacement used to expand the variable on the left of the '=' as well,
+		#    so it wrote "on=on" (or "off=off") and the key itself was lost from the settings file;
+		#  - UI_INTERFACE is a substring of GUI_INTERFACE, so an unanchored UI_INTERFACE expression
+		#    matches inside the GUI_INTERFACE line the previous expression has just produced and
+		#    rewrites it. Anchoring is what keeps the two apart.
 		sed -e "s/OH_MODE/$OH_MODE/g" -e "s/OH_LANGUAGE/$OH_LANGUAGE/g" -e "s&OH_DOC_DIR&../$OH_DOC_DIR&g" \
 		-e "s/DEMODATA=off/"DEMODATA=$DEMO_DATA"/g" -e "s/YES_OR_NO/$OH_SINGLE_USER/g" \
-		-e "s/PHOTO_DIR/$PHOTO_DIR_ESCAPED/g" -e "s/APISERVER=off/"APISERVER=$API_SERVER"/g" \
-		-e "s/GUI_INTERFACE=on/"$GUI_INTERFACE=$GUI_INTERFACE"/g" -e "s/UI_INTERFACE=off/"UI_INTERFACE=$UI_INTERFACE"/g" \
+		-e "s/PHOTO_DIR/$PHOTO_DIR_ESCAPED/g" \
+		-e "s/^APISERVER=off/APISERVER=$API_SERVER/" \
+		-e "s/^GUI_INTERFACE=on/GUI_INTERFACE=$GUI_INTERFACE/" \
+		-e "s/^UI_INTERFACE=off/UI_INTERFACE=$UI_INTERFACE/" \
 		./$OH_DIR/rsc/$OH_SETTINGS.dist > ./$OH_DIR/rsc/$OH_SETTINGS
 	fi
 
--- a/oh.ps1
+++ b/oh.ps1
@@ -1154,10 +1154,10 @@ function write_config_files {
 		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS").replace("DEMODATA=off","DEMODATA=$DEMO_DATA") | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
 		# set API_SERVER
 		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS").replace("APISERVER=off","APISERVER=$API_SERVER") | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
-		# set GUI INTERFACE
-		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS").replace("GUI_INTERFACE=on","GUI_INTERFACE=$GUI_INTERFACE") | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
+		# set GUI INTERFACE - anchored: UI_INTERFACE is a substring of GUI_INTERFACE
+		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS") -replace "^GUI_INTERFACE=on","GUI_INTERFACE=$GUI_INTERFACE" | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
 		# set UI INTERFACE
-		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS").replace("UI_INTERFACE=off","UI_INTERFACE=$UI_INTERFACE") | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
+		(Get-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS") -replace "^UI_INTERFACE=off","UI_INTERFACE=$UI_INTERFACE" | Set-Content "$OH_PATH/$OH_DIR/rsc/$OH_SETTINGS"
 	}
 
 	######## OH - Other settings setup
--- a/ohmac.sh
+++ b/ohmac.sh
@@ -266,20 +266,26 @@ function java_lib_setup {
 
 ###################################################################
 function write_api_config_file {
-	######## application.properties setup - OH API server	
-	SET_FILE=$OHDIR/rsc/$API_SETTINGS
-	if [ "$WRITE_CONFIG_FILES" = "on" ] || [ ! -f SET_FILE ]; then
-		[ -f $SET_FILE ] && mv -f $SET_FILE $SET_FILE.old		
-		# generate OH API token and save to settings file		
+	######## application.properties setup - OH API server
+	SET_FILE=./$OH_DIR/rsc/$API_SETTINGS
+	# The API templates only ship in the package that bundles the API server. Saying so beats
+	# writing an empty configuration file out of a template that is not there.
+	if [ ! -f "$SET_FILE.dist" ]; then
+		echo "Warning: $API_SETTINGS.dist not found, this package does not include the API server."
+		return
+	fi
+	if [ "$WRITE_CONFIG_FILES" = "on" ] || [ ! -f "$SET_FILE" ]; then
+		[ -f "$SET_FILE" ] && mv -f "$SET_FILE" "$SET_FILE.old"
+		# generate OH API token and save to settings file
 		JWT_TOKEN_SECRET=`LC_ALL=C tr -dc A-Za-z0-9 </dev/urandom | head -c 66`
 
-		echo "Writing OH API configuration file -> $API_SETTINGS..."
-		sed -i '' \
-			-e "s/JWT_TOKEN_SECRET/$JWT_TOKEN_SECRET/g" \
+		echo ">Writing OH API configuration file -> $API_SETTINGS..."
+		# read the template and write the copy, as oh.sh does: editing it in place would substitute
+		# the placeholders in the shipped file, and the next run would find nothing left to replace
+		sed -e "s/JWT_TOKEN_SECRET/$JWT_TOKEN_SECRET/g" \
 			-e "s/API_HOST:API_PORT/localhost:8080/g" \
 			-e "s&OH_API_PID&$OH_API_PID&g" \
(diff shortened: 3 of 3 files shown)


mwithi:
Thanks for completing the remaining macOS launcher fixes. Since `oh.sh` contains the most recent launcher configuration, could this PR align `ohmac.sh` with its current API/UI settings handling before merge?

In particular:

- read_settings in oh.sh also reads settings.properties and restores APISERVER, GUI_INTERFACE, UI_INTERFACE, mode, language and demo settings. In ohmac.sh, API_SERVER therefore starts as off on every run and UI_INTERFACE is never initialized, making the new start_ui path unreachable.
- EXPERT_MODE is initialized and the E) branch is implemented in oh.sh. Here the menu advertises -E and E is present in OPTSTRING, but the branch is commented out and EXPERT_MODE is undefined, so script_menu_advanced is also unreachable.
- write_api_config_file should be checked against the current placeholders handled by oh.sh (UI_HOST, UI_PORT, API_HOST, API_PORT, API_URL, and OH_API_PID). At present ohmac.sh uses the older combined API_HOST:API_PORT replacement and references OH_API_PID without defining it.

The actual API launch mechanism can remain platform/package-specific: oh.sh starts Tomcat, whereas directly launching the bundled executable JAR/WAR may be appropriate on macOS. The request is to align the persisted settings and variables with the newer script so the new API/UI functions are reachable and configured consistently.

giuseppesorge:
@mwithi done in `96b7199b` — all three points, plus one thing I hit on the way.

**`read_settings`** now reads `settings.properties` as `oh.sh` does and restores `APISERVER`, `GUI_INTERFACE`, `UI_INTERFACE`, mode, language and demo data. A `set_defaults` function reapplies the defaults afterwards for the keys the file does not carry — without it an absent key leaves its variable empty rather than at the value set at the top of the script, which is why `oh.sh` has the same block. `API_SERVER` therefore no longer starts as `off` on every run, and `UI_INTERFACE` is defined.

**`EXPERT_MODE`** is initialized and the `E` branch is implemented, so `script_menu_advanced` is reachable. I also added the `U` option and put it in `OPTSTRING` and in the advanced menu: without it there was still no way to turn `UI_INTERFACE` on from this script, so `start_ui` would have stayed unreachable even with the settings file read.

**`write_api_config_file`** now applies the same substitution list as `oh.sh` (`OH_API_PID`, `UI_HOST`, `UI_PORT`, `API_HOST`, `API_PORT`, `API_URL`) instead of the older combined `API_HOST:API_PORT`. That combined form left the `UI_HOST` and `UI_PORT` placeholders of `cors.allowed.origins` verbatim in the generated file. `OH_API_PID` is now declared — empty, as in `oh.sh`, where the assignment is commented out and the template resolves `spring.pid.file` to an empty value.

Kept platform specific, as you allowed: the launch mechanism, and `OH_UI_URL`. `oh.sh` appends `/$OH_UI_PROD` because it deploys the UI as a Tomcat webapp of that name; here the bundled Spring Boot artifact is started directly and the shipped `application.properties.dist` sets `server.servlet.context-path=/`, so the UI answers on the root path.

---

**One thing to flag in `oh.sh`.** While copying the settings-file substitutions I found that line 1019 does not do what it looks like:

```
-e "s/GUI_INTERFACE=on/"$GUI_INTERFACE=$GUI_INTERFACE"/g"
```

`$GUI_INTERFACE` is expanded on the left of the `=` as well, so the replacement is `on=on` (or `off=off`) and the key name is lost. `settings.properties` ends up with a line `on=on` and no `GUI_INTERFACE` at all — in every case, not only when the value changes.

There is a second, subtler one: `UI_INTERFACE` is a substring of `GUI_INTERFACE`. Even with the expansion fixed, the unanchored `s/UI_INTERFACE=off/…/g` matches inside the `GUI_INTERFACE=off` line the previous expression has just produced and rewrites it. Anchoring the three interface expressions at the start of the line fixes both:

```
-e "s/^APISERVER=off/APISERVER=$API_SERVER/"
-e "s/^GUI_INTERFACE=on/GUI_INTERFACE=$GUI_INTERFACE/"
-e "s/^UI_INTERFACE=off/UI_INTERFACE=$UI_INTERFACE/"
```

Verified on the shipped `settings.properties.dist` across the three combinations (`-U`, default, API only). That is what this PR uses. I have not touched `oh.sh` — happy to open a separate ticket for it if you want it fixed there too.

**Checks run:** with a package tree carrying the shipped `.dist` templates — settings written with `-U` come back as `APISERVER=on`, `GUI_INTERFACE=off`, `UI_INTERFACE=on` and survive a re-read, so both `start_api_server` and `start_ui` are reached; the generated `application.properties` has no placeholder left and the `.dist` template is untouched; `-E` and `-U` toggle both ways and the advanced menu lists what the script actually implements.

giuseppesorge:
Also here, `ebfa014e`: the bounded start/stop readiness check after `brew services`, asked for in the review of #2245.

It lands on this PR rather than that one because #2245 does not touch `ohmac.sh`, and two open pull requests editing the same file would leave a conflict for whichever is merged second.

`brew services start` returns when launchd has accepted the job, not when the server is ready, and the next thing the script does is run a mysql client — refused on a cold start, for a reason that has nothing to do with the installation. `brew services stop` has the mirror problem, with the port still held when the next start is attempted. The datadir fix from #2245 does not apply here, as you said: Homebrew owns the storage and the `OH_VERSION` check already covers it.

Same shape as `oh.sh`, probing with bash's `/dev/tcp` rather than `nc`, which is not part of a stock macOS either. Exercised against a real listener: port closed → exits on the timeout with code 2; listener up → returns immediately; stop-wait with the listener still up → warns and returns; listener closed → returns at once.

giuseppesorge:
Two more things here, `d7ff96ab` and `c8fd18e5`.

**The `oh.sh` bug I flagged above is fixed in this PR** rather than in a separate ticket — it belongs to the same alignment work, and the hunk does not overlap the `oh.sh` changes in #2245 (checked with `git merge-tree`: the two branches merge clean either way round).

Both problems are addressed: the three interface expressions are anchored at the start of the line and name the key on both sides, so `GUI_INTERFACE` is no longer replaced by `on=on`, and the `UI_INTERFACE` expression can no longer rewrite the `GUI_INTERFACE` line it has just been handed. Exercised against the shipped `settings.properties.dist` in the three combinations that actually occur — default, `-A`, `-U` — each of which now writes the three keys with their own names and the expected values.

**Rebased onto `develop` after #2244 was merged.** That merge touched `write_config_files` in `ohmac.sh`, which is the same function this branch changes, so the PR had gone to conflicting. Resolved keeping both sides: the `copy_config_file` calls for the remaining settings files that came with OP-1447, and the anchored substitutions from here. The whole check set was re-run on the merged file — defaults applied when the settings file is absent, `-U` persisted and read back so `start_api_server` and `start_ui` are reached, no placeholder left in the generated `application.properties` with the `.dist` untouched, `-E`/`-U` toggling both ways, and the four database wait cases.

The PR is mergeable again and still touches only `oh.sh` and `ohmac.sh`.

oh.sh
```diff
@@ -1013,10 +1013,19 @@ function write_config_files {
 	if [ "$WRITE_CONFIG_FILES" = "on" ] || [ ! -f ./$OH_DIR/rsc/$OH_SETTINGS ]; then
 		[ -f ./$OH_DIR/rsc/$OH_SETTINGS ] && mv -f ./$OH_DIR/rsc/$OH_SETTINGS ./$OH_DIR/rsc/$OH_SETTINGS.old
 		echo "Writing OH configuration file -> $OH_SETTINGS..."
+		# The three interface expressions are anchored at the start of the line, and name the key on
+		# both sides of the substitution:
+		#  - the GUI_INTERFACE replacement used to expand the variable on the left of the '=' as well,
+		#    so it wrote "on=on" (or "off=off") and the key itself was lost from the settings file;
+		#  - UI_INTERFACE is a substring of GUI_INTERFACE, so an unanchored UI_INTERFACE expression
+		#    matches inside the GUI_INTERFACE line the previous expression has just produced and
+		#    rewrites it. Anchoring is what keeps the two apart.
```
mwithi: Maybe commenting every single change is not needed, here the improvement is clear and straightforward. Let's keep the comments only for what is really hard to spot. Thanks!

```suggestion
```

oh.sh
```diff
@@ -1013,10 +1013,19 @@ function write_config_files {
...
 		sed -e "s/OH_MODE/$OH_MODE/g" -e "s/OH_LANGUAGE/$OH_LANGUAGE/g" -e "s&OH_DOC_DIR&../$OH_DOC_DIR&g" \
 		-e "s/DEMODATA=off/"DEMODATA=$DEMO_DATA"/g" -e "s/YES_OR_NO/$OH_SINGLE_USER/g" \
-		-e "s/PHOTO_DIR/$PHOTO_DIR_ESCAPED/g" -e "s/APISERVER=off/"APISERVER=$API_SERVER"/g" \
-		-e "s/GUI_INTERFACE=on/"$GUI_INTERFACE=$GUI_INTERFACE"/g" -e "s/UI_INTERFACE=off/"UI_INTERFACE=$UI_INTERFACE"/g" \
+		-e "s/PHOTO_DIR/$PHOTO_DIR_ESCAPED/g" \
+		-e "s/^APISERVER=off/APISERVER=$API_SERVER/" \
+		-e "s/^GUI_INTERFACE=on/GUI_INTERFACE=$GUI_INTERFACE/" \
+		-e "s/^UI_INTERFACE=off/UI_INTERFACE=$UI_INTERFACE/" \
```
mwithi, line 1022: should this improvement be done also on oh.ps1 script?

oh.sh
giuseppesorge: Done — the block is down to the one line that is not readable off the code:

```
# anchored: UI_INTERFACE is a substring of GUI_INTERFACE
```

I kept that one because it is the part that bites silently: without the `^`, the `UI_INTERFACE` expression matches inside the `GUI_INTERFACE` line it has just been handed, and nothing about the result looks wrong. The rest — what the anchors do, what the old expansion produced — is visible in the diff, and it is gone. Say the word and I will drop this line too.

Taken as general guidance for the rest of these branches, not just here.

oh.sh
giuseppesorge: Good catch — yes, half of it applies, in `85135f1f`.

The expansion mistake is **not** in `oh.ps1`: it names the key literally on both sides, so nothing writes `on=on` there.

The substring collision **is**. `String.replace` is a plain substring replacement, so once `GUI_INTERFACE` has been set to `off` the line reads `GUI_INTERFACE=off`, and the `UI_INTERFACE=off` replacement on the next line matches inside it and rewrites it. Run against the block as it stands in the file, under PowerShell:

```
=== oh.ps1 as it is ===
  API=off GUI=on  UI=off  ->  APISERVER=off GUI_INTERFACE=on UI_INTERFACE=off
  API=on  GUI=off UI=on   ->  APISERVER=on  GUI_INTERFACE=on UI_INTERFACE=on     <-- wrong
=== anchored ===
  API=on  GUI=off UI=on   ->  APISERVER=on  GUI_INTERFACE=off UI_INTERFACE=on
```

So on Windows, enabling the web interface left `GUI_INTERFACE=on` in `settings.properties`: the Swing client stayed enabled for a user who had asked for the web UI instead. Both replacements are now anchored regular expressions, verified in the three combinations that occur (default, `-A`, `-U`).

It lands on this PR next to the `oh.sh` twin rather than on #2245, even though that branch also touches `oh.ps1` — different part of the file, and I checked with `git merge-tree` that the two branches still merge clean either way round.

giuseppesorge:
@mwithi the `oh.ps1` fix is now confirmed on real Windows, under the shell the launcher actually ships for.

Windows runner (image `windows-2025-vs2026`), **Windows PowerShell 5.1**, running the nine substitution lines taken out of `oh.ps1` as they stand — not a reconstruction of them:

```
=== PowerShell 5.1.26100.32995 (Desktop) ===

--- 1. oh.ps1 parses ---
  OK - no syntax errors

--- 2. settings substitutions, running the real lines from oh.ps1 ---
  lines taken from oh.ps1 (rows 1197-1213): 9
  OK   API=off GUI=on  UI=off  ->  APISERVER=off GUI_INTERFACE=on  UI_INTERFACE=off
  OK   API=on  GUI=off UI=on   ->  APISERVER=on  GUI_INTERFACE=off UI_INTERFACE=on
  OK   API=on  GUI=on  UI=off  ->  APISERVER=on  GUI_INTERFACE=on  UI_INTERFACE=off

=== failures: 0 ===
```

Identical under PowerShell 7.6.3.

The middle row is the one that was wrong: before the change it wrote `GUI_INTERFACE=on` there, so enabling the web interface left the Swing client enabled as well. It now writes `off`, on Windows, on 5.1.

Worth noting for the merge order: the check needed both this branch and #2245, since `database_port_open` lives there and the settings block here. I merged the two on a scratch branch to run it and **they merge clean** — a second confirmation on top of the `git merge-tree` I mentioned earlier.

giuseppesorge:
And the `ohmac.sh` side is now checked on **macOS** too, on a clean `macos-latest` machine with the released `multiarch-client` package, rather than only on my own Mac:

```
--- 2. the settings file is read back, so the API server and the web UI are reachable ---
  OK   no settings file -> defaults off/on/off
  written: APISERVER=on GUI_INTERFACE=off UI_INTERFACE=on
  OK   read back on/off/on: start_api_server and start_ui are reachable
  OK   the GUI_INTERFACE key survived

--- 3. the API configuration ---
  OK   no API template -> warns and returns, writing nothing
  OK   no sed/cp error leaked

--- 4. -E and -U toggle, and the advanced menu lists what the script implements ---
  OK   -E turns expert mode on
  OK   -U turns the web UI on and the Swing GUI off
  OK   the advanced menu is reachable and lists -U

--- 5. the database waits are bounded ---
  OK   a port that never opens gives up after 3s with code 2
```

The third block is worth a word: the client package ships no API server, and `write_api_config_file` is reached there all the same. Before this branch that produced `sed:` and `cp:` errors against a path at the filesystem root — the check now asserts it says so and returns, writing nothing and leaking no shell error.

With the Linux and Windows runs, the three platforms this launcher family ships on are all covered by a run now.

mwithi (approved):
LGTM. Thanks a lot!

PR #2258, opened 2026-08-27
OP-1460 Run Open Hospital from the directory its own paths are resolved against

giuseppesorge:
## [OP-1460](https://openhospital.atlassian.net/browse/OP-1460)

Found while testing the v1.15.1 pre-release, so this targets `release_1_15_1`. Rebuilt on that base rather than cherry-picked from the develop branch it was first written on.

On macOS no report can be generated. The viewer never opens and the application log records:

```
net.sf.jasperreports.engine.JRException: java.io.FileNotFoundException: rpt_base/examslist.jasper
```

Reports are looked up **relative to the working directory** — `PrintManager.java:80` builds `new File("rpt_base/" + filename + ".jasper")`, and `JasperReportsManager` declares `RPT_BASE = "rpt_base"` — while in the package they live in `oh/rpt_base`. `oh.sh` knows it, and its `start_gui` opens with

```
cd "$OH_PATH/$OH_DIR" # workaround for hard coded paths
```

`ohmac.sh` never had it, so the application ran from the package root. The API server fails the same way, since no launcher starts it from inside `oh`: `/reports/diseases-list` and `/reports/exams-list` answer 500 with the same `FileNotFoundException`.

A second defect lives in the same lines. `$OH_PATH` was already used at line 72 to build `OH_PATH_ESCAPED`, but never set, so `dicom.properties` was written with

```
dicom.storage.filesystem=/data/dicom_storage
```

at the filesystem root, and every run printed an empty `OH_PATH is set to`.

## What changed

* `OH_PATH` is derived from the script location, the way `oh.sh` derives it, and can still be set from outside;
* everything handed to the JVM — classpath, native library path, java binary, API artifact — is absolute, so it survives the change of directory;
* `start_gui` and `start_api_server` change into `$OH_DIR` and return afterwards; the log files keep their current location;
* the jar pattern is expanded by the loop instead of going through `$DIRLIBS`, and the java invocations quote what they pass: the absolute form puts the installation directory into the classpath, which the relative one never carried, so a path containing a space would otherwise have dropped every jar.

## Verified

Running this branch's `ohmac.sh` on the released v1.15.1 multiarch client package:

* the JVM's working directory is `<package>/oh` — before the change, the package root;
* Exams Report renders: JasperViewer opens on *Page 1 of 5*. Before, no viewer appeared and the log carried the exception above;
* `dicom.storage.filesystem` is written inside the installation;
* the launcher prints the real path in `OH_PATH is set to ...`;
* on a copy of the package under a path containing a space: 159 jars on the classpath and the report still renders.

The same report was generated from the Linux portable package for comparison, where `oh.sh` already worked.

Not a regression of 1.15.1: `ohmac.sh` has never carried the workaround.

[OP-1460]: https://openhospital.atlassian.net/browse/OP-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ohmac.sh
+++ b/ohmac.sh
@@ -30,6 +30,15 @@ PHOTO_DIR="data/photo"
 BACKUP_DIR="data/dump"
 
 OH_DIR="oh"
+# Open Hospital resolves part of its own paths against the working directory - the reports are
+# looked up as rpt_base/... - so the application has to run from inside $OH_DIR, and every path
+# handed to the JVM has to be absolute to survive that. oh.sh carries the same workaround in its
+# own start_gui. The script already used $OH_PATH without ever setting it, so the DICOM storage
+# path was written as /data/dicom_storage, at the filesystem root.
+# Derived from the script location, as oh.sh does, and overridable the same way.
+if [ -z "${OH_PATH+x}" ]; then
+	OH_PATH=$(cd "$(dirname "$0")" && pwd)
+fi
 OH_DOC_DIR="doc"
 CONF_DIR="data/conf"
 DATA_DIR="data/db"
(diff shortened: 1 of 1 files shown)


giuseppesorge:
Re-running the checks: the build failed resolving `OH-core:1.15.1`, not on this change.

mizzioisf (approved):
LGMT (besides the comments, will work on that aftwerwars), can merge.

PR #2264, opened 2026-08-31
Updates for 1.15.1 release

mizzioisf:
- External software (JRE - MariaDB - Tomcat) update
- AI-generated content review and test
- Some other various fixes

### DIFF
--- a/oh.ps1
+++ b/oh.ps1
@@ -232,7 +232,7 @@ $script:EXPERT_MODE="off"
 
 ######## MariaDB/MySQL Software
 # MariaDB version
-$script:MYSQL_VERSION="10.6.23"
+$script:MYSQL_VERSION="10.6.28"
 $script:MYSQL32_VERSION="10.6.5"
 
 ######## define system and software architecture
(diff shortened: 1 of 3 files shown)


oh.sh
```diff
@@ -784,13 +784,11 @@ function initialize_database {
...
 
+
 ###################################################################
-# Whether the database is accepting connections on its TCP port.
-#
-# The port is probed with bash's own /dev/tcp redirection rather than with `nc`, which is not part
-# of the package and is absent from many minimal installations: there the probe never succeeded, so
-# the wait loop below span forever and the launcher hung with no message and no way to tell why.
 function database_port_open {
+	# Check if the database serveris accepting connections on its TCP port.
```
mwithi:
```suggestion
	# Check if the database server is accepting connections on its TCP port.
```

oh.ps1
```diff
@@ -300,7 +300,7 @@ function script_menu {
 	#
 	Write-Host " ------------------------------------------------------------------------"
 	Write-Host "|                                                                        |"
-	Write-Host "|                Open Hospital - v$OH_VERSION                                 |"
+	Write-Host "|                   Open Hospital - v$OH_VERSION                              |"
```
mwithi:
```suggestion
	Write-Host "|                   Open Hospital - v$OH_VERSION                         |"
```

oh.sh
```diff
@@ -228,7 +228,7 @@ function script_menu {
 	# show help / user options
 	echo " ------------------------------------------------------------------------"
 	echo "|                                                                        |"
-	echo "|                Open Hospital - v$OH_VERSION                                 |"
+	echo "|                   Open Hospital - v$OH_VERSION                              |"
```
dbmalkovsky, line 231: Does this need the same change as above?

mwithi: yeah, probably:


```suggestion
	echo "|                   Open Hospital - v$OH_VERSION                         |"
```

oh.sh
mizzioisf: yep ok will fix it

oh.sh
mwithi: deleted other suggestion, we have to consider the `$OH_VERSION` variable replacement that eventually will align the right border

mizzioisf: Not really, it is ok once the $OH_VERSION variable is filled with the current value :-)

PR #1684, opened 2023-02-04
OP-1026 add missing anamnesis related property strings

dbmalkovsky:
See OP-1026

PR #2266, opened 2026-09-01
Merge back after release 1.15.1

mwithi:


mizzioisf (approved):
LGTM !

dbmalkovsky (commented):
LGTM other than my question about the version

giuseppesorge:
One thing that passes silently here, because it is not a conflict: the merge back re-adds the 38 files under `rpt_base/en/`, which develop had dropped in 3dcc68a6 ("OP-214 Remove "en" folder as Transifex takes it as secondary source folder"). Today `develop` has 0 files there and `release_1_15_1` has 38, and all 38 show up as *added* in this diff.

Besides the Transifex reason, it also changes which bundle English reports read. In `JasperReportsManager.addReportBundleParameter` the order is `rpt_base/<lang>/<report>.properties` → `rpt_base/<report>_<lang>.properties` → `rpt_base/<report>.properties` (the last one being the fallback the core merge back brings in), so with the folder back in place the `en/` copy wins over the root file for English. 30 of the 38 are byte-identical to the root file on develop, but 8 are the older copies — the ones that matter:

- `patient_opd_chart` — `en/` has no `Id`, `Type`, `Date`, `Time`, `Diagnosis`, `SecondDiagnosis`, `ThirdDiagnosis`, `ClinicalRemarks`
- `ProductLedger` — no `of`, no `Page`
- `patient_clinical_sheet_ver3` — no `OPD`, and `Report.from` reads "report from" instead of "Report from"
- `InventoryWard` — `WardInventoryID` reads "WARD INVENTORY ID" instead of "Ward INVENTORY ID"
- the remaining 4 differ only in comments/blank lines

The root `rpt_base/*.properties` files are identical on the two branches, so removing `rpt_base/en/` from this merge would not lose any string. I can push that commit here if you like — or leave it as is, if the folder is meant to come back on develop.

giuseppesorge:
I pushed the removal as a branch in case it is useful: `giuseppesorge:remove-rpt-base-en` (48aa7778, on top of 7eb912aa). It deletes the 38 files and nothing else — every one of them has a root `rpt_base/*.properties` counterpart on this branch, so no string is lost.

One detail I had not checked when I wrote the comment above: `patient_clinical_sheet_ver3.jrxml` uses `$R{OPD}`, and the `en/` copy is the only bundle without that key. So dropping the folder also fixes that label in English on the release branch, not just the Transifex setup on develop.

I can open it as a PR against `release_1_15_1` (the way #1684 went in on core) if you prefer, or you can cherry-pick the commit — your call.

dbmalkovsky (approved):
LGTM

PR #2270, opened 2026-09-14
OP-1468 Align the macOS launcher with the JRE the other launchers install

giuseppesorge:
Fixes [OP-1468](https://openhospital.atlassian.net/browse/OP-1468).

Each launcher pins the JRE it downloads from the Azul CDN. "Updates for 1.15.1 release" (8218230d) bumped that pin in `oh.bat`, `oh.ps1` and `oh.sh`; `ohmac.sh` was not part of that change, so the released 1.15.1 packages install a different Java runtime on macOS than on the other two platforms.

From `OpenHospital-v1.15.1-multiarch-client.zip`, where all four launchers ship side by side and share one `oh/lib`:

| launcher | `JAVA_DISTRO` |
| --- | --- |
| `oh.sh` | `zulu17.68.203-ca-fx-jre17.0.20.1-linux_$JAVA_PACKAGE_ARCH` |
| `oh.ps1` | `zulu17.68.203-ca-fx-jre17.0.20.1-win_$JAVA_PACKAGE_ARCH` |
| `ohmac.sh` | `zulu17.60.17-ca-jre17.0.16-macosx_aarch64` |

`MYSQL_VERSION` and `TOMCAT_VERSION` were bumped in the same commit but do not apply here: on macOS the database comes from Homebrew and there is no bundled Tomcat. The JRE is the one pin the three launchers share.

Nothing is broken today - the older archive still resolves, so a first run on macOS installs a working runtime, just 17.0.16 instead of 17.0.20.1.

## Why this is two lines and not one

Zulu's macOS archives changed shape between the two versions:

* `zulu17.60.17-ca-jre17.0.16-macosx_aarch64` extracts a flat directory holding `bin/java`;
* `zulu17.68.203-ca-fx-jre17.0.20.1-macosx_aarch64` extracts an `.app`-style bundle that keeps it under `Contents/Home/bin/java`. The same is true of the non-`fx` build, so it is the version that changed, not the flavour.

`java_check` built that path by hand, so bumping `JAVA_DISTRO` alone would have left the launcher pointing at a path the new archive does not have, and it would have stopped at "Error: no usable Java found". `bundled_java_bin` looks for both shapes, which also means the next bump does not trip over this again.

## On the `-fx` flavour

The runtime the other two launchers now install carries JavaFX, which the codebase never references. This change matches it so the three stay on one runtime; dropping `-fx` from all three would be a smaller download, but that is a deliberate decision for all platforms rather than something to slip in here - happy to follow up separately if you would like it.

## Tested

On macOS 15 (Apple Silicon), against the released `OpenHospital-v1.15.1-multiarch-client.zip` with these changes applied to its launcher:

* with the current code, `$JAVA_DIR/bin/java` does not exist in the new archive - reproduced the failure the bump alone would have caused;
* with this change the path resolves to `Contents/Home/bin/java` and the JRE reports `17.0.20.1`;
* launching the shipped application with it, using the launcher's own classpath, `--add-opens` flags and `java.library.path`, starts Spring Boot on Java 17.0.20.1 and stops only at `DatabaseConfig` because a client package has no `database.properties` - no `UnsupportedClassVersionError`, all classes load.

Linux and Windows are untouched by this change.

[OP-1468]: https://openhospital.atlassian.net/browse/OP-1468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Closing: opened ahead of the discussion, not needed as a change right now. Sorry for the noise.

giuseppesorge:
Reopening at Alessandro's go-ahead. Unchanged from before: same commit, and develop has not moved since (still 821674ef).

mizzioisf (approved):
Should be ok (I don't have the possibility to test it)

mwithi (approved):

PR #1145, opened 2022-03-07
Updates for 1.11.3

mwithi:
- Various fixes and updates for CLIENT mode
- Better error checking
- Updated zulu JAVA JRE version
- Updated MariaDB version
- Needs testing

Signed-off-by: @mizzioisf

PR #2271, opened 2026-09-14
OP-1469 Check that the launchers pin the same third-party versions

giuseppesorge:
Fixes [OP-1469](https://openhospital.atlassian.net/browse/OP-1469).

> Stacked on #2270 - its commit is the first of the two here, so please merge #2270 first. Without it this check fails, which is the point: it is red on `develop` as it stands today.

`oh.sh`, `oh.ps1` and `ohmac.sh` each pin the JRE, MariaDB and Tomcat they download, so a release update has to touch every one of them. When one is missed nothing reports it - the packages are built and published with the launchers disagreeing, and it only surfaces when somebody runs that platform.

It has happened twice:

* OP-1438 - `ohmac.sh` stayed on a Java 11 JRE for a whole release cycle after the project moved to 17;
* OP-1468 - `oh.sh` and `oh.ps1` moved to JRE 17.0.20.1 in the 1.15.1 release update while `ohmac.sh` stayed on 17.0.16.

`.github/workflows/maven.yml` is the only workflow and it builds and tests the Java code, so nothing looks at the launchers at all.

## What it compares

* `JAVA_DISTRO` across the three, on the version part only - the archive name ends with the platform (`-linux_x64`, `-win_x64`, `-macosx_aarch64`), so the same runtime on three platforms counts as a match;
* `JAVA_URL` across the same three;
* `MYSQL_VERSION` and `TOMCAT_VERSION` between `oh.sh` and `oh.ps1`.

Left out on purpose, both noted in the script so the next reader does not have to rediscover it:

* `ohmac.sh` has no MariaDB or Tomcat pin - Homebrew provides the database and there is no bundled Tomcat;
* `oh.bat` only forwards to `oh.ps1` and pins nothing of its own;
* `MYSQL32_VERSION` has differed between `oh.sh` (`10.5.28`) and `oh.ps1` (`10.6.5`) since "Updates for 1.11.3" (#1145), so comparing it would report a difference that is meant to be there. If those two are actually supposed to match, that is worth its own look and I would add it to the check.

A renamed variable or a changed naming scheme fails the check instead of passing silently, so it cannot quietly stop checking.

## Why a separate job

It needs neither the JDK nor a companion core checkout, so it still reports while `build` is waiting on a dependent pull request in `openhospital-core`. It adds a few seconds.

## Tested

Run locally against five trees:

| tree | result |
| --- | --- |
| `develop` as it is today | fails, naming `ohmac.sh` and printing all three values |
| with #2270 applied | passes, all four pins reported |
| `JAVA_DISTRO` renamed in one launcher | fails - "has the variable been renamed?" |
| a non-Zulu archive name | fails - "the naming scheme has changed, so this check needs updating too" |
| `TOMCAT_VERSION` changed in `oh.sh` only | fails on Tomcat, the other three still pass |

This touches a shared CI file, so if you would rather have it as its own workflow, keep it out of `maven.yml`, or not have it at all, say the word and I will rework or drop it.

[OP-1469]: https://openhospital.atlassian.net/browse/OP-1469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Closing: opened ahead of the discussion, not needed as a change right now. Sorry for the noise.

giuseppesorge:
Reopening at Alessandro's go-ahead. Still stacked on #2270 — that one first.

mizzioisf:
I haven't the possibility to test it, but LGTM.

mwithi (approved):

PR #2272, opened 2026-09-15
OP-1470 Remove the duplicated database_port_open from oh.sh

giuseppesorge:
Fixes [OP-1470](https://openhospital.atlassian.net/browse/OP-1470).

`oh.sh` on `develop` declares `database_port_open` twice, at lines 789-793 and 795-802, with identical bodies:

```
(exec 3<>/dev/tcp/$DATABASE_SERVER/$DATABASE_PORT) > /dev/null 2>&1
```

Only the comment differs - one copy carries a short one inside the function, the other the explanatory block above it.

## How it got there

The function came in with OP-1443 (#2245) and reached the two branches with its comment in different places, so git saw two independent blocks rather than one moved comment:

| commit | date | definitions |
| --- | --- | --- |
| `8218230d` "Updates for 1.15.1 release" | 01/09 | 1 |
| `9c71d5c7` "Merge branch 'develop' into release_1_15_1" | 02/09 | 2 |
| `6ef58c9d` "Merge back after release 1.15.1" | 08/09 | 2, now on `develop` |

**The released package was never affected**: the `v1.15.1` tag has a single definition. The duplicate appeared on the branch the day after the tag was cut.

## What this changes

Nothing at runtime. Bash keeps the last definition and the two are the same, so the launcher behaves identically either way - this removes dead code that a reader would otherwise have to diff by hand to be sure of.

The copy that survives is the one carrying the reason the port is probed with bash's own `/dev/tcp` rather than with `nc`.

## Checked

* one definition left, `bash -n oh.sh` clean;
* both call sites intact - the `until` at line 829 and the `while` at 967 after the removal;
* the resulting region is **byte-identical to the same region at the `v1.15.1` tag**, i.e. this restores exactly the shape the file had before the duplication, including the blank line the extra block had brought with it;
* no other function is declared twice in `oh.sh`, and neither `ohmac.sh` nor `oh.ps1` has any.

## Follow-up

A rule for "function declared more than once in the same launcher" would have caught this. It belongs with the launcher consistency check in #2271, and is better added after this merges so the check does not start out red on `develop`. Happy to push it there once this is in.

[OP-1470]: https://openhospital.atlassian.net/browse/OP-1470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mizzioisf (approved):
LGTM, let's merge

mwithi (approved):

PR #2273, opened 2026-09-15
OP-1471 Remove rpt_base/en again after the merge back re-added it

giuseppesorge:
Fixes [OP-1471](https://openhospital.atlassian.net/browse/OP-1471).

OP-214 removed `rpt_base/en` on purpose - Transifex treats a language subfolder named after the source language as a secondary source. The merge back after release 1.15.1 brought all 38 files back: `develop` had dropped them, the release branch had not, and git kept them.

## It is not only a Transifex matter

`addReportBundleParameter` in core resolves a report's bundle in this order:

1. `rpt_base/<language>/<report>.properties`
2. `rpt_base/<report>_<language>.properties`
3. `rpt_base/<report>.properties`

So with the folder back, for English the `en/` copies shadow the root files - and those copies are older than the ones that have been kept up to date.

## Nothing is lost

All 38 files have a counterpart in `rpt_base/`, and **no key present in an `en/` copy is missing from its root file** - checked key by key, not by eye. Thirty are identical to the root file, four differ only in comments or blank lines, four carry real differences.

## What actually changes for a reader

| report | key | effect |
| --- | --- | --- |
| `patient_clinical_sheet_ver3` | `OPD` | the jrxml uses `$R{OPD}` and the `en/` copy is the only bundle without it, so the label is unresolved today and starts working again |
| `patient_clinical_sheet_ver3` | `Report.from` | "report from" → "Report from" |
| `InventoryWard` | `WardInventoryID` | "WARD INVENTORY ID" → "Ward INVENTORY ID" |
| `ProductLedger` | `Page`, `of` | none - the jrxml does not reference them |
| `patient_opd_chart` | eight keys | none - the jrxml does not reference them |

**The one worth your opinion is `WardInventoryID`.** It is the only change that arguably goes the wrong way: the root file's "Ward INVENTORY ID" mixes cases where the `en/` copy had it all caps. If you prefer the all-caps form, the right fix is to correct the root file rather than to keep the folder - happy to add that here.

I said "eleven keys recovered" when I first reported this on #2266; that was counting keys rather than effects. Only `OPD` is actually referenced by a jrxml - the other ten are dead either way.

## Noticed while checking, out of scope

Two pre-existing mismatches, present with or without this change: `patient_adm_chart.jrxml` uses `$R{Disease Out}` while the key is `DiseaseOut`, and `patient_clinical_sheet_ver3.jrxml` uses `$R{To}` while the key is `to`. Both render the key name instead of the label. Happy to file them if useful.

[OP-1471]: https://openhospital.atlassian.net/browse/OP-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


giuseppesorge:
Filed the two pre-existing mismatches I mentioned above as [OP-1472](https://openhospital.atlassian.net/browse/OP-1472), so they do not get lost with this change.

Turned out to be a bit more interesting than "the label is broken": both jrxml files declare `whenResourceMissingType="Key"`, so the key name is printed instead of failing - and since the key names are `Disease Out` and `To`, the labels read correctly in English and stay English in every other language. `DiseaseOut` and `to` are translated in it/fr/es/de, the keys actually referenced are defined in no bundle at all, in any of the fourteen language folders.

Left out of this PR on purpose: fixing them means editing the jrxml, and both ship a committed `.jasper` that would need rebuilding - the question OP-1403 is about.

[OP-1472]: https://openhospital.atlassian.net/browse/OP-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

mwithi:
Hi @giuseppesorge could you please add to this PR also `rpt_stat/en/` and `rpt_extra/en/`? Thanks!

giuseppesorge:
Done — 16 files from `rpt_stat/en`, 17 from `rpt_extra/en` (ff59f773). Checked key by key: nothing is lost, and 15 English labels that today print their own key name start resolving from the root files (`jTitle` on OH001-OH008, `AllMaterials`/`TestsDone` on OH007, `Male`/`Female` on OH008).

One exception: `BillsReportMonthlyWard.jasper` declares `resourceBundle="BillsReportMonthly"`, so with no `en/` folder English falls back to a bundle that never had the three ward keys and the subtitle would silently print "Month.Summary.per.Ward". Added them in 1340661c — happy to drop that commit if you'd rather keep the PR to the two folders; the proper fix is the jrxml plus a `.jasper` rebuild (OP-1403).

Correction to my comment above on `rpt_base`: four more labels change than the three I listed — `DetailsInvoice`, `PATIENT_BILLS_SUMMARY`, `Total` → `TOTAL`, `Pat.N` → `PAT. N.`

mwithi (approved):
LGTM. Thanks a lot!

